Issue #14, opened 2024-02-16
Add `show` method for `CharTable`


fingolfin:
Right now:
```
julia> g=genchartab("GL2")
[...]

PR #37, opened 2024-03-05
Add show method for character tables

fingolfin:
Progress towards #14

Right now Green function tables are printed identically to "actual" generic character tables. Simply because there seems to be no way to actually distinguish the two after they have been loaded.


SoongNoonien:
Ok, your merged faster than I could look at the changes as your requested a review :-). But I think this looks fine.

Issue #16, opened 2024-02-16
Getting the number of classes / characters?


fingolfin:
It would be great if there was a way to compute the number of classes / characters (i.e. not just the number of class / character *types*) as a polynomial in `q`.

SoongNoonien:
So this can be closed I think.

PR #40, opened 2024-03-05
added `irrchars` and `classes` methods for tables

SoongNoonien:
@fingolfin Is this what you meant in #16?

### DIFF
--- a/src/Show.jl
+++ b/src/Show.jl
@@ -7,6 +7,7 @@ export classtypes
 export irrchartypes
 export nrchars
 export nrclasses
+export nrirrchars
 export nrparams
 export order
 export param
@@ -69,6 +70,26 @@ function irrchartypes(t::Table)
 end
 
 """
+    nrirrchars(t::Table)
+
+Return the number of irreducible characters of table `t`.
+
+For example this excludes characters created with [`tensor!`](@ref) or [`lincomb!`](@ref).
+
+# Examples
+```jldoctest
+julia> g=genchartab(\"GL2\");
+
+julia> nrirrchars(g)
+q^2 - 1
+
+```
+"""
+function nrirrchars(t::Table)
+	return sum(nrchars.(Ref(t), 1:irrchartypes(t)))
+end
+
+"""
     classtypes(t::Table)
 
 Return the number of conjugacy class types of table `t`.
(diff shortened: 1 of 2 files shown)


SoongNoonien:
The names are not ideal, see #41. But I don't know better names right now.

src/Show.jl
```diff
@@ -49,6 +49,26 @@ function irrchartypes(t::Table)
...
+```jldoctest
+julia> g=genchartab(\"GL2\");
+
+julia> irrchars(g)
+q^2 - 1
+
+```
+"""
+function irrchars(t::Table)
+	return sum(nrchars.(Ref(t), 1:irrchartypes(t)))
```
fingolfin, line 69: TIL one can use `Ref` like that in a broadcast, nice :-)

src/Show.jl
```diff
@@ -49,6 +49,26 @@ function irrchartypes(t::Table)
 	return t.table.original_m
 end
 
+"""
+    irrchars(t::Table)
```
fingolfin, line 53: I would just call this `nrchars` as well.

And `irrchartypes` should be nirrchartypes` (or rather in OSCAR style: `n_irrchartypes` -- or even `number_of_irreducible_character_types` or whatever... )

src/Show.jl
```diff
@@ -67,6 +87,24 @@ function classtypes(t::Table)
 	size(t.table, 2)
 end
 
+"""
+    classes(t::Table)
```
fingolfin, line 91: As above I'd just call this `nrclasses` as well

And eventually rename both to `n_classes`, and `classtypes` to `n_classtypes` -- but please not in this PR, this needs to be carefully coordinated.

fingolfin:
I've resolved the conflicts and updated this to look like I meant.

But I am now not so sure this is a good idea. Isn't it true that some of the character/class types may not be disjoint? At least that's how I read the CHEVIE paper from 1996. ...

Anyway I don't need this right now, I can try to talk with e.g. Gunter Malle about this (I want to do that anyway for a bunch of things in this code)

SoongNoonien:
> But I am now not so sure this is a good idea. Isn't it true that some of the character/class types may not be disjoint? At least that's how I read the CHEVIE paper from 1996. ...

I don't know, I thought that this is not the case if we consider the parameter exceptions which is done here.

src/Show.jl
```diff
@@ -68,6 +68,26 @@ function irrchartypes(t::Table)
...
+# Examples
+```jldoctest
+julia> g=genchartab(\"GL2\");
+
+julia> nrirrchars(g)
+q^2 - 1
+
+```
+"""
+function nrirrchars(t::Table)
```
SoongNoonien, line 87: Is this function exported?

fingolfin: Oops! Now it is.

SoongNoonien:
Should we wait with the merge until it is clear if we are counting some characters twice?

SoongNoonien:
I've talked to Gunter Malle today and he thinks this should be fine, so now characters/classes are be counted multiple times.

PR #44, opened 2024-03-05
Remove some redundancy from the data tables

fingolfin:
This mostly started out by me trying to verify that `ParameterSubstitution` are not actually ever "preset" to a value in the tables. The other steps then were just similar cleanup -- and as a bonus now one can `git grep ParameterException` and get something useful 

(I guess it also saves some little bit of disk space but that was not a motivation)


SoongNoonien (approved):
This looks cleaner than before.

Issue #26, opened 2024-02-29
Document `setcongruence`


fingolfin:

fingolfin:
I also really wonder  how it can be used. I'd like to take e.g. handle the case where `q` is even. In CHEVIE one can substitute `q` by `2*q` to do that.

Or to tell the program that `(q-1)/3` is an integer one can substitute 3q+1 for q.

To quote the CHEVIE paper:
```
> nr:=‘SL3.1‘[-2,0];
nr := A2204
> print(setCongruence.nr);
proc() global qQ,q; qQ := ’qQ’; q := 3*qQ+1; NULL end
> print(unsetCongruence.nr);
proc() global q,qQ; q := ’q’; qQ := 1/3*q-1/3; NULL end
```
Note that it has these two substitution rules so that it can apply and then *unapply* the change as needed; my understanding this is because some code needs the "original" q (after all, `3q+1` in general won't be a prime power (and even if, it would be coprime to `q`, so ....)

Maybe in this package I don't even need `setcongruence` for this and should just substitute `q`? But

SoongNoonien:
> Maybe in this package I don't even need setcongruence for this and should just substitute q?

But isn't that just exactly what we currently do? Or is there a better way to substitute `q` than to evaluate the polynomials?

fingolfin:
OK, so how do I do it? I am trying to reproduce examples from the CHEVIE paper. E.g.:
```
julia> T = genchartab("SL3.n1");

julia> tensor!(T,2,2)
9

```
I would now like to specialize the table or the new character or whatever to "even q". In the paper this is done via `setcongruence` which substitutes `2q` (or rather `2Q` for a new variable `Q`) for `q`. But I couldn't figure out how to do that here:

"Our" `setcongruence` is not called anywhere (well only in one `simplify` function but that doesn't make it clearer to me) and there are no examples. It seems from its signature that it is mostly intended for use on values, not on a whole table or a table row (aka a "character type")?


The other option I found is `speccharparam!` but I couldn't get it to do this either:
```
julia> q, (a, b, m, n) = params(T)
(q, (a, b, m, n))

julia> speccharparam!(T, 9, q, 2*q)
ERROR: MethodError: no method matching speccharparam!(::GenericCharacterTables.CharTable{…}, ::Int64, ::Nemo.QQPolyRingElem, ::Nemo.QQPolyRingElem)

Closest candidates are:
[...]

PR #45, opened 2024-03-05
Add setcongruence for a table

fingolfin:
Goes a step towards what has been discussed in issue #26


fingolfin:
I don't claim this is perfect but I need it, so... we can revise it as needed later.

PR #48, opened 2024-03-15
Add myself to authors list

fingolfin:
While the vast bulk of the package of course was done by Andreas Leim and Martin Wagner, it was a project under my supervision, and currently I am doing the ongoing development.

I'd be more than happy to add a paragraph to the manual that explains the genesis and who did what, if you'd like to provide one, @SoongNoonien (I don't know Andreas Leims' GitHub account otherwise I'd ping him here, too).


SoongNoonien:
> I'd be more than happy to add a paragraph to the manual that explains the genesis and who did what, if you'd like to provide one, @SoongNoonien

I'm not sure if it can be clearly separated who did what. I know Andreas did most of the table conversion code and the initial tests for latex output.

> I don't know Andreas Leims' GitHub account otherwise I'd ping him here, too

I don't think he has one. But I'll ask him.

PR #50, opened 2024-04-16
fixed some indentations

SoongNoonien:
I know that this is not the most important thing right now but I noticed it while looking through the code. @fingolfin it seems like most of the inconsistencies were introduced by you. Do you like me to use spaces rather than tabs? I'm not sure if Oscar has a clear preference here.


fingolfin:
We are not very strict about this in Oscar but we do have a `.JuliaFormatter.toml` file there which defines our preferences and which you could use together with JuliaFormatter

fingolfin (approved):

PR #51, opened 2024-04-25
reworked `nesum`

SoongNoonien:


src/SumProc.jl
```diff
@@ -107,36 +108,45 @@ function nesum(a::CycloFrac{T}, var::Int64, lower::Int64, upper::Union{Int64,T},
...
-				sum+=summand
+				sum+=geometric_sum
 			else
-				sum+=simplify(summand, c, cinv)
+				sum+=simplify(geometric_sum, c, cinv)
 			end
-			if !(ishalf(root.argument) && isunitfraction(cl))
+			# If `var_coeff` evaluates to an integer `e2p(var_coeff)` evaluates to one and thus `e2p(var_coeff)-1` to zero.
+			# So in this case the closed formular for the geometric sum doesn't hold.
+			if !(ishalf(var_coeff) && isunitfraction(var_coeff))  # TODO make this more general to skip more exceptions?
```
fingolfin, line 137: We just discussed this code and after reading the code if `ishalf` and `isunitfraction` it seems as if this check says: "if `var_coeff` is 1 or 1/2 then DO NOT add parameter exceptions" -- but if I understand @SoongNoonien right, then it should really only exclude this for "1/2" ??   I guess the function `ishalf` is a bit confusing (and perhaps needs a better name)

SoongNoonien: Ok, I've checked again and it seems like `var_coeff` can't be 1. `summand.argument` is reduced and doesn’t have any integral parts. So something like `summand.argument == 1//q*i+j` (with `var == j` this would mean `var_coeff == 1`) is not possible this would always be simplified to `1//q*i`. This is why we added the comment in line 33. But this comment is rather confusing.

src/SumProc.jl
```diff
@@ -107,36 +108,45 @@ function nesum(a::CycloFrac{T}, var::Int64, lower::Int64, upper::Union{Int64,T},
 		sum, exc = nesum(a, var, 0, upper, congruence)
 		return (sum-nesum(a, var, 0, lower-1), exc)
```
fingolfin, line 109: Unrelated to this PR, but:

Is this code actually correct? Wouldn't the `nesum` call produce two return values, and so we get an error here? And what about the `congruence` argument?

How about this:
```suggestion
        sum2, exc2 = nesum(a, var, 0, lower-1, congruence)
		return sum-sum2, union(exc, exc2)
```

Perhaps not in this PR but in another, and perhaps you could add a test case or two for this?

SoongNoonien: Oh, you are correct. This wouldn't work.

Issue #24, opened 2024-02-29
Ensure parameter exceptions of e.g. scalar products are printed nicer?


fingolfin:
Right now we have e.g. this:
```
julia> T = genchartab("SL3.n1")

julia> scalar(T,1,4)
(0, Set(GenericCharacterTables.ParameterException{Nemo.QQPolyRingElem}[2//(q - 1)*n2 ∈ ℤ]))
```
It already gets a lot better if the involved types are imported into `Main`:
```
julia> import GenericCharacterTables: ParameterException, QQPolyRingElem

julia> scalar(T,1,4)
(0, Set(ParameterException{QQPolyRingElem}[2//(q - 1)*n2 ∈ ℤ]))
```
But this doesn't really help with larger examples:
```
julia> scalar(T,4,6)
(0, Set(ParameterException{QQPolyRingElem}[1//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*m2 - 1//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*n1 - 1//(q - 1)*m2 + 1//(q - 1)*n2 ∈ ℤ, 2//(q - 1)*n1 + 1//(q - 1)*m2 - 2//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*n1 + 2//(q - 1)*m2 - 1//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*n1 - 1//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*n1 + 1//(q - 1)*m2 ∈ ℤ, 1//(q - 1)*n1 - 1//(q - 1)*m2 + 2//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*m2 ∈ ℤ, 2//(q - 1)*n1 + 1//(q - 1)*m2 + 1//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*n1 - 1//(q - 1)*m2 - 1//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*n1 + 1//(q - 1)*m2 - 1//(q - 1)*n2 ∈ ℤ, 2//(q - 1)*n1 - 2//(q - 1)*m2 + 1//(q - 1)*n2 ∈ ℤ, 1//(q - 1)*n1 - 1//(q - 1)*m2 ∈ ℤ, 1//(q - 1)*n1 + 1//(q - 1)*n2 ∈ ℤ]))
```
[...]

PR #58, opened 2024-05-10
move `ParameterExceptions` to `CycloFrac` to improve printing

SoongNoonien:
It seems like I broke something here. Running the example from #24 now returns less exceptions. Maybe they are still sufficient but it would be nice to see what is going on. And I have to updated the tests...

### DIFF
--- a/src/Arith.jl
+++ b/src/Arith.jl
@@ -154,12 +154,9 @@ struct CycloFrac{T} <: Cyclotomic{T}
 			elseif numerator == denominator
 				o=one(denominator)
 				return new{T}(o, o, exceptions)
-			else
-				return new{T}(numerator, denominator, exceptions)
 			end
-		else
-			return new{T}(numerator, denominator, exceptions)
 		end
+		return new{T}(numerator, denominator, exceptions)
 	end
 end
 
@@ -198,17 +195,18 @@ function Base.show(io::IO, z::CycloFrac)
 	end
 	if isone(length(z.numerator.summands))
 		if isone(length(z.denominator.summands))
-			print(io, "$(z.numerator)//$(z.denominator)$(exceptions)")
+			print(io, "$(z.numerator)//$(z.denominator)")
 		else
-			print(io, "$(z.numerator)//($(z.denominator))$(exceptions)")
+			print(io, "$(z.numerator)//($(z.denominator))")
 		end
 	else
 		if isone(length(z.denominator.summands))
-			print(io, "($(z.numerator))//$(z.denominator)$(exceptions)")
+			print(io, "($(z.numerator))//$(z.denominator)")
 		else
-			print(io, "($(z.numerator))//($(z.denominator))$(exceptions)")
+			print(io, "($(z.numerator))//($(z.denominator))")
 		end
 	end
+	print(io, exceptions)
 end
 Base.show(io::IO, m::MIME{Symbol("text/latex")}, z::CycloFrac) = print(io, "\\frac{$(repr("text/latex", z.numerator))}{$(repr("text/latex",z.denominator))}")
 Base.isone(x::CycloFrac) = isone(x.numerator) && isone(x.denominator)
--- a/src/Arith.jl
+++ b/src/Arith.jl
@@ -210,7 +210,18 @@ function Base.show(io::IO, z::CycloFrac)
 end
 Base.show(io::IO, m::MIME{Symbol("text/latex")}, z::CycloFrac) = print(io, "\\frac{$(repr("text/latex", z.numerator))}{$(repr("text/latex",z.denominator))}")
 Base.isone(x::CycloFrac) = isone(x.numerator) && isone(x.denominator)
-Base.iszero(x::CycloFrac) = iszero(x.numerator)  # TODO deal with exceptions?
+
+"""
+    iszero(x::CycloFrac; ignore_exceptions::Bool=false)
+
+Return if `x` is zero. If `ignore_exceptions` is true then the exceptions of `x` will not be considered.
+"""
+function Base.iszero(x::CycloFrac; ignore_exceptions::Bool=false)
+	if ignore_exceptions
+		return iszero(x.numerator)
+	end
+	return iszero(x.numerator) && isempty(x.exceptions)
+end
 
 # operators
 """
@@ -283,7 +294,7 @@ Base.:-(x::Cyclotomic{<:NfPoly}, y::Union{Int64, Rational{Int64}}) = x-(y*one(x)
 Base.:-(y::Cyclotomic) = (-1)*y
 
 Base.:(==)(x::CycloSum, y::CycloSum) = iszero(x-y)
-Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero(x-y)  # TODO deal with exceptions?
+Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero((x-y).numerator)
 
 function Base.:*(x::Cyclo, y::Cyclo)
 	if isone(x)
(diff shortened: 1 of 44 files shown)


SoongNoonien:
Ok, the tests are fixed now as well as the problem with the examples from #24.

src/Arith.jl
```diff
@@ -149,39 +146,71 @@ Base.iszero(x::CycloSum) = iszero(x.summands[1])  # isone(length(x.summands)) is
...
-				return new{T}(numerator, denominator)
+				return new{T}(numerator, denominator, exceptions)
 			end
 		else
-			return new{T}(numerator, denominator)
+			return new{T}(numerator, denominator, exceptions)
 		end
 	end
 end
+function add_exception!(a::CycloFrac{T}, exception::ParameterException{T}) where T<:PolyRingElem
```
fingolfin, line 165:
```suggestion

function add_exception!(a::CycloFrac{T}, exception::ParameterException{T}) where T<:PolyRingElem
```

src/Arith.jl
```diff
@@ -149,39 +146,71 @@ Base.iszero(x::CycloSum) = iszero(x.summands[1])  # isone(length(x.summands)) is
...
+	push!(a.exceptions, exception)
+end
+"""
```
fingolfin, line 168:
```suggestion

"""
```

src/Arith.jl
```diff
@@ -149,39 +146,71 @@ Base.iszero(x::CycloSum) = iszero(x.summands[1])  # isone(length(x.summands)) is
...
+			print(io, "$(z.numerator)//($(z.denominator))$(exceptions)")
 		end
 	else
 		if isone(length(z.denominator.summands))
-			print(io, "($(z.numerator))//$(z.denominator)")
+			print(io, "($(z.numerator))//$(z.denominator)$(exceptions)")
 		else
-			print(io, "($(z.numerator))//($(z.denominator))")
+			print(io, "($(z.numerator))//($(z.denominator))$(exceptions)")
 		end
```
fingolfin, line 208: simpler patch: just do `println(io, exceptions)` at the end. That will also be more amenable to future changes to e.g. adopt the AbstractAlgebra/OSCAR printing modes (terse/oneline/supercompact, see [here](https://docs.oscar-system.org/dev/DeveloperDocumentation/printing_details/))

src/Arith.jl
```diff
@@ -149,39 +146,71 @@ Base.iszero(x::CycloSum) = iszero(x.summands[1])  # isone(length(x.summands)) is
...
 	end
 end
 Base.show(io::IO, m::MIME{Symbol("text/latex")}, z::CycloFrac) = print(io, "\\frac{$(repr("text/latex", z.numerator))}{$(repr("text/latex",z.denominator))}")
 Base.isone(x::CycloFrac) = isone(x.numerator) && isone(x.denominator)
-Base.iszero(x::CycloFrac) = iszero(x.numerator)
+Base.iszero(x::CycloFrac) = iszero(x.numerator)  # TODO deal with exceptions?
```
fingolfin, line 213: If this is meant to implement the ring interface then this should require `isempty(x.exceptions)` (otherwise `x` isn't neutral with respect to addition).

src/Arith.jl
```diff
@@ -149,39 +146,71 @@ Base.iszero(x::CycloSum) = iszero(x.summands[1])  # isone(length(x.summands)) is
...
-				return new{T}(o, o)
+				return new{T}(o, o, exceptions)
 			else
-				return new{T}(numerator, denominator)
+				return new{T}(numerator, denominator, exceptions)
 			end
 		else
-			return new{T}(numerator, denominator)
+			return new{T}(numerator, denominator, exceptions)
 		end
```
fingolfin, line 162: Possible code simplifcation
```suggestion
		end
		return new{T}(numerator, denominator, exceptions)
```
... then remove the now empty `else`... next you can remove lines 157/158 (i.e. prior `else return ...`

src/Arith.jl
```diff
@@ -254,7 +283,7 @@ Base.:-(x::Cyclotomic{<:NfPoly}, y::Union{Int64, Rational{Int64}}) = x-(y*one(x)
 Base.:-(y::Cyclotomic) = (-1)*y
 
 Base.:(==)(x::CycloSum, y::CycloSum) = iszero(x-y)
-Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero(x-y)
+Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero(x-y)  # TODO deal with exceptions?
```
fingolfin, line 286: If you change `iszero` to honor exceptions, this one is fine as well.
```suggestion
Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero(x-y)
```

src/Show.jl
```diff
@@ -81,7 +81,7 @@ For example this excludes characters created with [`tensor!`](@ref) or [`lincomb
 julia> g=genchartab(\"GL2\");
 
 julia> nrirrchars(g)
-q^2 - 1
+q^2 - 1//1
```
fingolfin, line 84: Could we omit the `//1` in this case?

src/Arith.jl
SoongNoonien: Currently it is not meant to implement the ring interface.

src/Show.jl
SoongNoonien: Yes, we could and I think we should eventually. I need to fix the whole printing of the `CycloFrac` type anyways...

src/Arith.jl
SoongNoonien: If I implemented your suggestion regarding `iszero` any `CycloFrac` with at least one exception wouldn't even be equal to itself anymore as far as I understand it.

src/Arith.jl
SoongNoonien: Shouldn't it be just `print` not `println`?

SoongNoonien: `print` seems to do the trick.

src/Arith.jl
fingolfin: Ah, true. 

For this and other applications it would be good to have a variant of is_zero which ignores the exceptions.

src/Arith.jl
```diff
@@ -178,10 +206,22 @@ function Base.show(io::IO, z::CycloFrac)
 			print(io, "($(z.numerator))//($(z.denominator))")
 		end
 	end
+	print(io, exceptions)
```
fingolfin, line 209: We can merge it, but ideally, I would restructure this:
```suggestion
    if !isempty(z.exceptions)
	    print(io, "\nWith exceptions:")
	    for exception in z.exceptions
			print(io, "\n  $(exception)")
		end
	end
```
Actually it could also be rewritten to use `join(io, z.exceptions, "\n  ")` I guess -- well and even better: make it use `AbstractAlgebra.Indent()` and `pretty(io)` etc. to realize the indentation. But all that could be in a future PR.

src/Arith.jl
```diff
@@ -254,7 +294,7 @@ Base.:-(x::Cyclotomic{<:NfPoly}, y::Union{Int64, Rational{Int64}}) = x-(y*one(x)
 Base.:-(y::Cyclotomic) = (-1)*y
 
 Base.:(==)(x::CycloSum, y::CycloSum) = iszero(x-y)
-Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero(x-y)
+Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero(x-y, ignore_exceptions=true)
```
fingolfin, line 297: .... or just do this?
```suggestion
Base.:(==)(x::CycloFrac, y::CycloFrac) = iszero((x-y).numerator)
```

src/Arith.jl
SoongNoonien: I'll do that in another PR.

src/Arith.jl
SoongNoonien: Yes, that's also fine.

PR #64, opened 2024-05-28
add `AbstractGenericCharacter` type

SoongNoonien:
I'm still working on this. Currently it's just a relatively useless duplication of data.

### DIFF
--- a/src/CharTable.jl
+++ b/src/CharTable.jl
@@ -1,30 +1,7 @@
 export genchartab, greenfuntab
 
 abstract type Table{T} end
-
-# This matrix type is needed to be able to extend the character tables by new character types calculated by the methods in ModifyTable.jl
-# https://discourse.julialang.org/t/silly-question-how-to-add-a-row-to-array/33601/10
-mutable struct ExtendableMatrix{T} <: AbstractMatrix{T}
-	m::Int
-	original_m::Int
-	n::Int
-	data::Vector{T}
-	function ExtendableMatrix{T}(m::Int, n::Int, data::Vector{T}) where T
-		new{T}(m,m,n,data)
-	end
-end
-Base.size(a::ExtendableMatrix) = (a.m, a.n)
-Base.getindex(a::ExtendableMatrix, i::Integer, j::Integer) = a.data[(Int(i)-1)*a.n + Int(j)]
-Base.setindex!(a::ExtendableMatrix, v, i::Integer, j::Integer) = setindex!(a.data, v, (Int(i)-1)*a.n + Int(j))
-ExtendableMatrix{T}(::UndefInitializer, m::Integer, n::Integer) where {T} = ExtendableMatrix{T}(m, n, Array{T}(undef, m*n))
-ExtendableMatrix(a::AbstractMatrix{T}) where {T} = copyto!(ExtendableMatrix{T}(undef, size(a)...), a)
-function Base.push!(a::ExtendableMatrix, row::AbstractVector)
-	a.n == length(row) || throw(DimensionMismatch("row size must match matrix"))
-	resize!(a.data, length(a) + a.n)
-	a.data[length(a)+1:length(a.data)] = row
-	a.m += 1
-	return a
-end
+abstract type AbstractGenericCharacter{T} end
 
 # This is the main generic character table type.
 # T is usually of the type NfPoly.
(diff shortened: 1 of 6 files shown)


src/CharTable.jl
```diff
@@ -36,21 +37,47 @@ end
...
 	classparamindex::Vector{Int64}  # Indices of the class parameters
 	charparamindex::Vector{Int64}  # Indices of the character parameters
 	classparams::Vector{Parameters{T}}  # Info about the parameters of each class type
-	charparams::Vector{Parameters{T}}  # Info about the parameters of each character type
+	charparams::Vector{Parameters{T}}  # Info about the parameters of each character type  # TODO remove
 	congruence::Union{Tuple{T, T}, Nothing}  # Congruence of the main parameter q (of T). q is congruent to congruence[1] mod congruence[2].
 	modulusring::PolyRing  # Ring of polynomials of type T used in table (also ring of modulus of Cyclotomics)
 	argumentring::Generic.UniversalPolyRing{Generic.FracFieldElem{T}, Generic.MPoly{Generic.FracFieldElem{T}}}  # Ring of argument of the Cyclotomics in table
 	information::String  # General info about the table
+	chars::Vector{<:Any}
```
fingolfin, line 55:
```suggestion
	chars::Vector{<:AbstractGenericCharacter{T}}
```

data/Greenfunctions/2A2/GU3.jl
```diff
@@ -1,5 +1,5 @@
 using ..GenericCharacterTables
-import ..GenericCharacterTables: ExtendableMatrix, SimpleCharTable
+import ..GenericCharacterTables: SimpleCharTable
```
fingolfin, line 2: This PR is getting a bit unwieldy. Here is a suggestion so that we can merge parts of it quickly now, and reduce this PR to its core: namely, let's put the changes which expunge `ExtendableMatrix` from the `data` files (without changing anything else) into their own (initial) commit of this PR, and then also open another PR with just that commit.

We can work on that together on Wednesday if you like, but you can also try it out now. It involves some mild `git` history rewriting. Let me sketch what I would do below (apologies if you already know all of this):
1. First make sure there are no uncommitted changes; also perhaps do a `git branch backup` to keep a pointer to your top commit just in case something goes wrong
2. now go to the root directory of your repository clone and do `git reset master` (this will remove all commits but keep your changes)
3. `git add data src/Convert/Convert.jl && git commit -m "Remove ExtendableMatrix references from data/"`
4. `git commit -m "add AbstractGenericCharacter type" src test`  -- commit rest of the changes
5. `git checkout -b rm-ExtendableMatrix`   -- switch to new branch
6. `git reset --hard HEAD^`  -- discard the last commit (it's still safely on your original branch)
7. edit `src/CharTable/jl` to add a `CharTable` constructor just like you do in this PR, only it just passes all arguments through, except for the second argument `table`, which is turned into `ExtendableMatrix(table)`
8. `git commit -m "Adjust CharTable constructor"`
9. create a pull request (`gh pr create` if you use the `gh` command line tool)
10. afterwards switch back to the original branch and rebase that onto the new one to simplify things: `git rebase rm-ExtendableMatrix` (depending on how and where you added the new constructor in step 7, this will either require resolving a merge conflict, or removing a now duplicate `CharTable` constructor.

fingolfin: (err, and obviously you also need to treat `SimpleCharTable` the same way by adding an auxiliary constructor for it, too

src/ModifyTable.jl
```diff
@@ -2,6 +2,36 @@ export tensor!, omega!, lincomb!, speccharparam!, specclassparam!
...
+function Base.:*(char1::GenericCharacter{T}, char2::GenericCharacter{T}) where T<:PolyRingElem
+	if char1.parent != char2.parent
+		throw(DomainError((char1.parent,char2.parent), "Tables do not match."))
+	end
+	t=char1.parent
+	new_char_degree=char1.degree*char2.degree
+	new_char_values=Vector{CycloSum{T}}(undef, classtypes(t))
+	for class in range(1, classtypes(t))
+		val1=shift_char_parameters(t, char1.values[class], 4)
+		val2=shift_char_parameters(t, char2.values[class], 5)
```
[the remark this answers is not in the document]
SoongNoonien: I agree that these look a bit magical. Every table has 6 predefined sets of variables. First the normal one used at construction time of the tables. Then there are 3 sets used in the Ortho.jl and finally there are the two sets used for tensor products. So we shift one character type 4 steps ahead to the 5th set of variables and the other character types 5 steps to the 6th and last set of variables. Possibly this should be written in a comment somewhere.

src/ModifyTable.jl
```diff
@@ -2,6 +2,36 @@ export tensor!, omega!, lincomb!, speccharparam!, specclassparam!
 
 # TODO deal with ParameterSubstitution, this is not done in the original implementation.
 
+function Base.:*(char1::GenericCharacter{T}, char2::GenericCharacter{T}) where T<:PolyRingElem
+	if char1.parent != char2.parent
```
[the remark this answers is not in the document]
SoongNoonien: This should be done now.

data/Greenfunctions/2A2/GU3.jl
SoongNoonien: I hope I got you right, but I think this should be done now.

src/ModifyTable.jl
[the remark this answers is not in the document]
fingolfin: Yes this in a comment would be great

fingolfin (approved):

PR #70, opened 2024-06-18
Character printing improvement

SoongNoonien:


src/Show.jl
```diff
@@ -24,9 +24,25 @@ export status
...
 import Oscar: pretty, Indent, Dedent
 
+"""
+    show(io::IO, t::Table)
+
+Display a summary of the generic character table `t`.
+
+# Examples
+```jldoctest
+julia> g=genchartab(\"GL2\")
```
fingolfin, line 34:
```suggestion
julia> g=genchartab("GL2")
```

src/Show.jl
```diff
@@ -24,9 +24,25 @@ export status
 
 import Oscar: pretty, Indent, Dedent
 
+"""
```
fingolfin, line 27: Hint: if you need to use e.g. `$` in a docstring, you can also do this:
```suggestion
@doc raw"""
```

But even without this it should be fine to use `"`

fingolfin (approved):

PR #74, opened 2024-06-25
implement OSCAR's 2+1 printing for tables and chars

SoongNoonien:


src/Show.jl
```diff
@@ -57,6 +61,13 @@ function Base.show(io::IO, t::Table)
 	end
 end
 
+function Base.show(io::IO, t::Table)
+	print(io, "Generic character table")
+	if !is_terse(io)
+		print(io, " ", t.importname)
```
fingolfin, line 67: I would ignore terse mode here and just always print the name

PR #76, opened 2024-07-02
Delete tables_info.txt

fingolfin:
do we still need this file?

### DIFF
--- a/tables_info.txt
+++ b/tables_info.txt
@@ -1,3 +1,0 @@
-"many" Exceptions:
-	2F4/*
-	3D4/*


SoongNoonien:
We never actually needed it, it think. But it reminds me that the tables in 2F4 and 3D4 have uncompleted parameter exceptions. In the original implementation they didn't bother to save all parameter exceptions and just saved a string saying `"many"`. We didn't implement this for the `ParameterException` type. While the parameter exceptions are just informational data which isn't used for any computation in the future this should change to get rid of the class sums and char sums.

SoongNoonien (approved):

PR #82, opened 2024-07-12
added basic ring interface for generic cyclotomics

SoongNoonien:
This is a first implementation of a ring interface which should replace `Cyclo` and `CycloSum` while supporting the congruence simplification which was previously done in `Congruence.jl`. I'm still having some problems with the promotion and conversion. Also I think the ring constructor should actually be named `generic_cyclotomic_ring` to be more in line with the new naming scheme of Oscar.

### DIFF
--- a/src/GenericCyclotomics.jl
+++ b/src/GenericCyclotomics.jl
@@ -1,0 +1,484 @@
+using Oscar.Random: Random, SamplerTrivial, GLOBAL_RNG
+using Oscar.RandomExtensions: RandomExtensions, Make2, AbstractRNG
+
+import Oscar.AbstractAlgebra: parent_type, elem_type, base_ring, base_ring_type, parent, is_domain_type,
+	is_exact_type, canonical_unit, isequal, divexact, zero!, mul!, add!, addeq!,
+	get_cached!, is_unit, characteristic, Ring, RingElem, expressify, evaluate
+import Oscar: pretty, Indent, Dedent, terse, is_terse
+import Oscar.AbstractAlgebra: normal_form
+
+import Base: show, +, -, *, ^, ==, inv, isone, iszero, one, zero, rand, deepcopy_internal, hash, conj
+
+import Compat
+
+export kempner, kempner_with_data, generic_cyclotomic_ring
+
+if Compat.pkgversion(Oscar.AbstractAlgebra) >= v"0.42.0"
+	const UPoly = Generic.UnivPoly{QQFieldElem}
+	const UPolyRing = Generic.UniversalPolyRing{QQFieldElem}
+else
+	const UPoly = Generic.UnivPoly{QQFieldElem, Generic.MPoly{QQFieldElem}}
+	const UPolyRing = Generic.UniversalPolyRing{QQFieldElem, Generic.MPoly{QQFieldElem}}
+end
+const UPolyFrac = Generic.FracFieldElem{UPoly}
+const UPolyFracRing = Generic.FracField{UPoly}
+
+@doc raw"""
+    kempner_with_data(m::Int64)
+
+Return the `m`-th Kempner number, the associated Kempner schemes and a factorization of `m`.
+
+# Examples
+```jldoctest
+julia> kempner_with_data(12)
+(4, Dict(2 => [1, 3], 3 => [1]), 1 * 2^2 * 3)
+
+```
+"""
+function kempner_with_data(m::Int64)
+	# factor m for later use
+	factorization=factor(m)
+
+	# compute m-th Kempner number by using the schemes introduced by Kempner
+	# these schemes can later be reused to compute reductions of m
+	schemes=Dict{Int64, Vector{Int64}}()
+	for (p,a) in factorization
+		multiple=p
+		exponent=1
+		schemes[p]=Int64[exponent]
+		while exponent < a
+			multiple+=p
+			exponent+=valuation(multiple, p)
+			push!(schemes[p], exponent)
+		end
+	end
+	if isone(m)
+		K_m=0
+	else
+		K_m=maximum((length(scheme)*p for (p,scheme) in schemes))
+	end
+
+	return (K_m, schemes, factorization)
+end
+
+@doc raw"""
+    kempner(m::Int64)
+
+Return the `m`-th Kempner number.
+
+# Examples
+```jldoctest
+julia> kempner(12)
+4
+
+```
+"""
+kempner(m::Int64) = kempner_with_data(m)[1]
+
+@doc raw"""
+    normal_form(f::AbstractAlgebra.Generic.UnivPoly{ZZRingElem, AbstractAlgebra.Generic.MPoly{ZZRingElem}}, m::Int64)
+
+Return a normal form of `f` modulo `m`, such that `normal_form(f,m)` is equal
+to `normal_form(g,m)` if and only if `f` and `g` are congruent modulo `m`.
+
+# Examples
+```jldoctest
+julia> R=universal_polynomial_ring(ZZ);
+
+julia> x=gen(R, :x);
+
+julia> normal_form(4*x^2,6)
+x^2 + 3*x
+
+julia> normal_form(4*x^2-(x^2+3*x),6)
+0
+
+julia> normal_form(4*x^9+x^7,12)
+x^3 + 4*x
+
+julia> normal_form(4*x^9+x^7-(x^3+4*x),12)
+0
+
+```
+"""
+function normal_form(f::AbstractAlgebra.Generic.UnivPoly{ZZRingElem, AbstractAlgebra.Generic.MPoly{ZZRingElem}}, m::Int64)
+	if m < 1
+		throw(DomainError(m, "A normal form for non-positive moduli is not defined!"))
+	end
+	R=parent(f)
+	if is_constant(f)
+		return R(mod(constant_coefficient(f),m))
+	end
+	# find the variable with highest index occurring in f
+	i=findlast(x -> degree(f,x)>0, gens(R))
+	x_i=R[i]
+
+	# save Kempner schemes to be reused to compute reductions of m
+	K_m,schemes,factorization=kempner_with_data(m)
+
+	# reduce by the K(m)-th rising factorial of x_i
+	S=rising_factorial(x_i, K_m)
+	f=mod(f,S)
+
+	for k in reverse(0:(K_m-1))
+		# compute the reduction of m by k by reusing the schemes
+		reduction=1
+		for (p,scheme) in schemes
+			q=div(k,p)
+			if iszero(q)
+				reduction*=p^factorization[p]
+			elseif q <= length(scheme)
+				exponent=factorization[p]-scheme[q]
+				if exponent > 0
+					reduction*=p^exponent
+				end
+			end
+		end
+
+		# reduce coefficient of x_i^k
+		S_ik=rising_factorial(x_i, k)
+		C_ik=coeff(f,[x_i],[k])
+		c=normal_form(C_ik, reduction)
+		f-=(C_ik-c)*S_ik
+	end
+	return f
+end
+
+strip_zeros!(f::Dict{UPolyFrac, UPoly}) = filter!(p->!iszero(p.second), f)
+
+mutable struct GenericCycloRing <: Ring
+	base_ring::UPolyRing
+	congruence::Union{Tuple{QQFieldElem, QQFieldElem}, Nothing}
+end
+
+mutable struct GenericCyclo <: RingElem
+	f::Dict{UPolyFrac, UPoly}
+	parent::GenericCycloRing
+end
+
+# Data type and parent object methods
+
+parent_type(::Type{GenericCyclo}) = GenericCycloRing
+
+elem_type(::Type{GenericCycloRing}) = GenericCyclo
+
+base_ring_type(::Type{GenericCycloRing}) = UPolyRing
+
+base_ring(R::GenericCycloRing) = R.base_ring::base_ring_type(R)
+
+parent(f::GenericCyclo) = f.parent
+
+is_domain_type(::Type{GenericCyclo}) = false
+
+is_exact_type(::Type{GenericCyclo}) = true
+
+function hash(x::GenericCyclo, h::UInt)
+	r = 0x65125ab8e0cd44ca  # TODO Use a different hash this is the one from AA's example.
+	return xor(r, hash(x.f, h))
+end
+
+function deepcopy_internal(x::GenericCyclo, dict::IdDict)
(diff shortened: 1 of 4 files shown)


src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,462 @@
...
+	for (p,a) in factorization
+		multiple=p
+		exponent=1
+		schemes[p]=Int64[exponent]
+		while exponent < a
+			multiple+=p
+			exponent+=valuation(multiple, p)
+			push!(schemes[p], exponent)
+		end
+	end
```
fingolfin, line 64: Maybe move this into a helper, e.g. `kempner_scheme` or so?

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,462 @@
+using Oscar.Random: Random, SamplerTrivial, GLOBAL_RNG
+using Oscar.RandomExtensions: RandomExtensions, Make2, AbstractRNG
+
+import Oscar.AbstractAlgebra: parent_type, elem_type, base_ring, parent, is_domain_type,
```
fingolfin, line 4:
```suggestion
import Oscar.AbstractAlgebra: parent_type, elem_type, base_ring, base_ring_type, parent, is_domain_type,
```

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,462 @@
...
+	is_exact_type, canonical_unit, isequal, divexact, zero!, mul!, add!, addeq!,
+	get_cached!, is_unit, characteristic, Ring, RingElem, expressify
+import Oscar: pretty, Indent, Dedent, terse, is_terse
+import Oscar.AbstractAlgebra: normal_form
+
+import Base: show, +, -, *, ^, ==, inv, isone, iszero, one, zero, rand, deepcopy_internal, hash
+
+const UPoly = Generic.UnivPoly{QQFieldElem, Generic.MPoly{QQFieldElem}}
```
fingolfin, line 12:
```suggestion
const UPoly = Generic.UnivPoly{QQFieldElem}
```

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,462 @@
...
+	end
+end
+
+const GenericCycloID = AbstractAlgebra.CacheDictType{Ring, GenericCycloRing}()
+
+mutable struct GenericCyclo <: RingElem
+	f::Dict{UPolyFrac, UPoly}
+	parent::GenericCycloRing
+
+	function GenericCyclo(f::Dict{UPolyFrac, UPoly})
```
fingolfin, line 125: I would suggest to make the parent a mandatory argument for the constructor, as in your case the parent always is constructed beforehand anyway.

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,462 @@
...
+		end
+	end
+	return f
+end
+
+mutable struct GenericCycloRing <: Ring
+	base_ring::UPolyRing
+	congruence::Union{Tuple{QQFieldElem, QQFieldElem}, Nothing}
+
+	function GenericCycloRing(R::UPolyRing, congruence::Union{Tuple{QQFieldElem, QQFieldElem}, Nothing}, cached::Bool)  # TODO
```
fingolfin, line 112: Remove the caching for now: as long as the congruence inside the GenericCycloRing can be changed by the user, this will lead to incorrect results

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,445 @@
...
+	end
+	# find the variable with highest index occurring in f
+	i=findlast(x -> degree(f,x)>0, gens(R))
+	x_i=R[i]
+
+	# factor m for later use
+	factorization=factor(m)
+
+	# compute m-th Kempner number by using the schemes introduced by Kempner
+	# these schemes can later be reused to compute reductions of m
```
fingolfin, line 53: Please add references! I.e. add the Kempner paper to the bib and then point to specific sections.

That would help anyone else trying to understand this code.

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,445 @@
...
+	f=mod(f,S)
+
+	for k in reverse(0:(K_m-1))
+		# compute the reduction of m by k by reusing the schemes
+		reduction=1
+		for (p,scheme) in schemes
+			factor=div(k,p)
+			if iszero(factor)
+				reduction*=p^factorization[p]
+			elseif factor <= length(scheme)
```
fingolfin, line 82: So what if `factor > length(scheme)` ? Is doing nothing correct then?

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,445 @@
...
+
+	# reduce by the K(m)-th rising factorial of x_i
+	S=rising_factorial(x_i, K_m)
+	f=mod(f,S)
+
+	for k in reverse(0:(K_m-1))
+		# compute the reduction of m by k by reusing the schemes
+		reduction=1
+		for (p,scheme) in schemes
+			factor=div(k,p)
```
fingolfin, line 79: is `p` a divisor of `k` at this point? Then I'd change this to
```suggestion
			factor = k / p
```
Or is it maybe not (I think this is the case)? Then I would rename `factor` to e.g. `quotient`, `quot`, `q`, ... as it isn't, well, a factor of `k`

src/GenericCyclotomics.jl
SoongNoonien: Yes, this is correct. If `factor > length(scheme)` then we only need to multiply `reduction` by one.

src/GenericCyclotomics.jl
SoongNoonien: > Or is it maybe not (I think this is the case)?

No, `p` is not necessarily a divisor of `k`.

> Then I would rename factor to e.g. quotient, quot, q, ... as it isn't, well, a factor of k

Yes, I can do this. I had troubles to find a good name for `factor`.

src/GenericCyclotomics.jl
SoongNoonien: When I do this I get the following when trying to construct a generic cyclotomic:
```
julia> S(1)
ERROR: MethodError: Cannot `convert` an object of type 
  AbstractAlgebra.Generic.FracFieldElem{AbstractAlgebra.Generic.UnivPoly{QQFieldElem, AbstractAlgebra.Generic.MPoly{QQFieldElem}}} to an object of type 
  AbstractAlgebra.Generic.FracFieldElem{AbstractAlgebra.Generic.UnivPoly{QQFieldElem, U} where U<:MPolyRingElem{QQFieldElem}}

Closest candidates are:
  convert(::Type{T}, ::T) where T
   @ Base Base.jl:84
  AbstractAlgebra.Generic.FracFieldElem{T}(::T, ::T) where T<:RingElem
   @ AbstractAlgebra ~/.julia/packages/AbstractAlgebra/88ogO/src/generic/GenericTypes.jl:915

Stacktrace:
 [1] setindex!(h::Dict{…}, v0::AbstractAlgebra.Generic.UnivPoly{…}, key0::AbstractAlgebra.Generic.FracFieldElem{…})
   @ Base ./dict.jl:367
 [2] Dict{AbstractAlgebra.Generic.FracFieldElem{…}, AbstractAlgebra.Generic.UnivPoly{…} where U<:MPolyRingElem{…}}(kv::Dict{AbstractAlgebra.Generic.FracFieldElem{…}, AbstractAlgebra.Generic.UnivPoly{…}})
   @ Base ./dict.jl:84
 [3] convert
   @ ./abstractdict.jl:568 [inlined]
 [4] GenericCharacterTables.GenericCyclo(f::Dict{AbstractAlgebra.Generic.FracFieldElem{…}, AbstractAlgebra.Generic.UnivPoly{…}}, parent::GenericCharacterTables.GenericCycloRing)
   @ GenericCharacterTables ~/Archiv/Computer/Programme/GenericCharacterTables.jl/src/GenericCyclotomics.jl:140
 [5] (::GenericCharacterTables.GenericCycloRing)(c::Int64)
   @ GenericCharacterTables ~/Archiv/Computer/Programme/GenericCharacterTables.jl/src/GenericCyclotomics.jl:341
 [6] top-level scope
   @ REPL[20]:1
Some type information was truncated. Use `show(err)` to see complete types.

```

thofma: Then we probably have to do the version dance. Since `pkgversion` is not available on julia 1.6, we could do the following hack to detect the new AA version:
```
@static isdefined(AbstractAlgebra.Solve, :MatrixNormalFormTrait)
  const UPoly = Generic.UnivPoly{QQFieldElem, Generic.MPoly{QQFieldElem}}
else
  const UPoly = Generic.UnivPoly{QQFieldElem}
```

SoongNoonien: > Since pkgversion is not available on julia 1.6, ...

I think we don't need 1.6 support. We're not even testing against 1.6.

> or something like that,...

Sigh... Ok, I hoped we could get around this.

thofma: If 1.6 is not supported, the compat version should be adjusted.

SoongNoonien: I think I've found a solution which should work with 1.6 as well. It's the alternative solution we had for AA 0.42 https://github.com/oscar-system/GenericCharacterTables.jl/pull/85.

SoongNoonien: @fingolfin What do you think about julia 1.6 support and testing?

SoongNoonien:
I think this should be ready now. Of course it won't do very much at this point but adapting the code to this interface will require modifying most files which would make this pull request a huge mess. The next step should be porting `CycloFrac` (ideally under a different name) to this ring and away from `CycloSum`. After that I should be able to replace all occurrences of the old `Cyclotomics` with their new counterparts.

src/GenericCyclotomics.jl
fingolfin: I think I disabled testing on Julia 1.6 in CI at some point due to problems. I'll see what the current status is and if we can restore support, see PR #128.

Since OSCAR currently supports Julia >= 1.6, as 1.6 is the Julia LTS release. So it would be kinda nice if we could also support it. But only if we can do it with reasonable effort.

Note that Julia 1.10 is expected to become the next Julia LTS release pretty soon. And then OSCAR might follow suite after some time, which would also resolve this (but of course then we should definitely adjust `Project.toml` to require a newer Julia.

SoongNoonien: Ok, so this PR seems to be compatible with 1.6.

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,489 @@
...
+	end
+	return f
+end
+
+function strip_zeros!(f::Dict{UPolyFrac, UPoly})
+	for (key,value) in f
+		if iszero(value)
+			delete!(f, key)
+		end
+	end
```
fingolfin, line 143: More efficient:
```suggestion
	filter!(p->!iszero(p.second), f)
```

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,489 @@
...
+end
+
+function show(io::IO, R::GenericCycloRing)
+	print(io, "Generic cyclotomic ring")
+	if !is_terse(io)
+		print(io, " over ", base_ring(base_ring(R)))
+	end
+end
+
+function show(io::IO, x::GenericCyclo)  # TODO
```
fingolfin, line 225: Maybe the TODO here is about "change this to use expressify system?

fingolfin (approved):
Left a bunch of comments. I'll leave it to you to decide whether to merge this now and address them later, or address some now, and some later, or whatever.

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,489 @@
...
+base_ring(R::GenericCycloRing) = R.base_ring::base_ring_type(R)
+
+parent(f::GenericCyclo) = f.parent
+
+is_domain_type(::Type{GenericCyclo}) = false
+
+is_exact_type(::Type{GenericCyclo}) = true
+
+function hash(x::GenericCyclo, h::UInt)
+	r = 0x65125ab8e0cd44ca  # TODO
```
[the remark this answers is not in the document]
SoongNoonien: This is the exact same hash used in the AA example. I thought this should be a different one. But I'm not sure which is the best way to create a new one.

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,489 @@
...
+	end
+	if isone(m)
+		K_m=0
+	else
+		K_m=maximum((length(scheme)*p for (p,scheme) in schemes))
+	end
+
+	# optionally return the schemes and the factorization
+	if extras
+		return (K_m, schemes, factorization)
```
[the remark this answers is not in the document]
SoongNoonien: I don't know how often `kempner` will be called in a usual computation. But I expect it to be run not very often and mostly for small values (less than three decimal digits). We'll see how the performance changes when I finally port the code to use this ring.

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,489 @@
...
+	end
+end
+
+# Unary operations
+
+function -(x::GenericCyclo)
+	f=deepcopy(x.f)
+	for (key,value) in f
+		f[key]=-value
+	end
```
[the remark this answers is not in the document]
SoongNoonien: Ah, nice! I had some problems with using `map` on dicts. But I think `map!(-, values(f))` would be even shorter.

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,489 @@
...
+
+*(x::UPoly, y::GenericCyclo) = parent(y)(x)*y
+
+*(x::GenericCyclo, y::UPoly) = y*x
+
+# Comparison
+
+function ==(x::GenericCyclo, y::GenericCyclo)
+	parent(x) != parent(y) && error("Incompatible rings")
+	return iszero(x-y)
```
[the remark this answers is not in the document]
SoongNoonien: Yes, unfortunately there are different representations for the same generic cyclotomic. This can be improved by using Zumbroich bases for the constant part of the exponents. But I only found out about them relatively recently. I'm not sure if they completely fix the problem though. Maybe there are some nasty cases...

src/GenericCyclotomics.jl
```diff
@@ -0,0 +1,489 @@
...
+	for (key,value) in x.f
+		f[-key]=value
+	end
+	return parent(x)(f)
+end
+
+# Binary operations
+
+function +(x::GenericCyclo, y::GenericCyclo)
+	parent(x) != parent(y) && error("Incompatible rings")
```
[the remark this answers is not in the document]
SoongNoonien: Ah, nice!

PR #84, opened 2024-07-17
Update UnivPolyRing usage

lgoettgens:
An approach to fixing the issue described in https://github.com/Nemocas/AbstractAlgebra.jl/pull/1748#issuecomment-2227892485.

If this both works on the master and the release-0.2 branch, releasing this on both branches would make https://github.com/Nemocas/AbstractAlgebra.jl/pull/1748 non-breaking.

### DIFF
--- a/src/GenericCharacterTables.jl
+++ b/src/GenericCharacterTables.jl
@@ -19,7 +19,7 @@ module GenericCharacterTables
 using Oscar
 import Oscar.AbstractAlgebra.Generic
 
-const FracPoly{T} = Generic.UnivPoly{Generic.FracFieldElem{T}, Generic.MPoly{Generic.FracFieldElem{T}}} where T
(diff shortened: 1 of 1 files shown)


fingolfin:
I didn't see this. Yes this would work, at the cost of some performance when using AA before 0.42, so in particular with Oscar 1.0 and 1.1. But that's negligible.

But I disagree that this would render the change in AA nonbreaking: the CI booktests for Oscar 1.1.x would still be broken, as they hardcode version 0.2.0 of this package.

lgoettgens:
> But I disagree that this would render the change in AA nonbreaking: the CI booktests for Oscar 1.1.x would still be broken, as they hardcode version 0.2.0 of this package.

Yeah. The time line would be to merge either this or #85, release it, update the Oscar master book tests to use a newer or the latest version of GCT and finally bumping the AA compat of Oscar (where we included the change in a breaking release)

PR #85, opened 2024-07-18
Prepare for AA 0.42.0

fingolfin:
Note that the new code won't be used for now because Oscar does not yet support the new AA.


fingolfin:
This is in anticipation of https://github.com/Nemocas/AbstractAlgebra.jl/pull/1748 so maybe we wait for that to be merged before we merge this PR here.

CC @lgoettgens @thofma @SoongNoonien

SoongNoonien:
This looks good to me.

SoongNoonien:
But I'm still surprised why #84 seems to be working too. Apparently it doesn't need the case distinction.

fingolfin:
Closed in favor of PR #84

PR #86, opened 2024-07-22
Update TagBot.yml

lgoettgens:
This is how it is suggested to be used in https://github.com/JuliaRegistries/TagBot?tab=readme-ov-file#setup

I imagine that this is a reason that TagBot doesn't do proper releases on this repository, but only tags versions.

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -4,6 +4,22 @@ on:
     types:
       - created
   workflow_dispatch:
+    inputs:
+      lookback:
+        default: 3
+permissions:
+  actions: read
+  checks: read
+  contents: write
+  deployments: read
+  issues: read
+  discussions: read
+  packages: read
+  pages: read
+  pull-requests: read
+  repository-projects: read
+  security-events: read
+  statuses: read
 jobs:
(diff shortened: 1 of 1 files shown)


fingolfin:
> TagBot doesn't do proper releases on this repository, but only tags versions

Not sure what you mean by that? They do have releases, but the way GitHub Action work is via git refs, including tags that stand in for "1.x" (i.e. `v1`) and so on.

Anyway this seems to be irrelevant for this PR

fingolfin:
Oh wait I totally misunderstood you: what you meant is that TagBot failed to create releases on *this* repository. Yeah, true. But I think that was the permissions in the project settings, which I just adjusted.

fingolfin:
I.e. the picture just below the section in the TagBot README that you linked to

fingolfin:
(My mistake was that I read the two paragraphs in your description subsequently; so after looking at the TagBot README link pointing at the TagBot repository, I read "this repository" as "the TagBot repository". Which is "obviously nonsense", but as so often with misunderstandings, the "obvious" part only is "obvious" after you understood what you misunderstood *sigh*)

Issue #88, opened 2024-07-23
Issues with changed return value of `scalar`


lgoettgens:
At some point before 0.3.0, the return value of `scalar` was changed from 2 to 1 objects.
Some code (e.g. https://github.com/oscar-system/GenericCharacterTables.jl/blob/36c227ece743a6ab3a86c23a0304115ae396d53f/src/Show.jl#L570) still relies on the old behaviour.

Additionally, this change broke https://github.com/oscar-system/Oscar.jl/blob/1870bfd798c8b2edb3a452b9c543924fb5f09219/test/book/cornerstones/groups/genchar.jlcon#L57 in the OSCAR book.

How do we proceed with this?

SoongNoonien:
@fingolfin What do you think about this? I think `print_decomposition` can easily be adapted to the new return value. All the exception checking is now done within `CycloFrac` itself.

fingolfin:
The idea always was that for the book example as printed on paper, we will require GCT 0.2.x; but newer breaking releases such as 0.3.0 can and will change interfaces to be more comfortable. As such I am not overly concerned about such breakage -- instead of adjusting GCT to allow the code example work unchanged, we'll update the book website with additional versions of the example: "This is the GenericCharacterTable example as printed in the book. Click here for a revised version using a much nicer interface for GCT, as announced in the book"

Of course I don't want to break compatibility gratuitously, but here there really is no good alternative.


(Of course the book proofs are still not here, despite Springer's promises; so hypothetically I could even still adjust the book (!) to match GCT 0.3.x behavior...)

fingolfin:
Of course `print_decomposition` should be adjusted to work, and we should add a test to make sure that it works. But the book example also calls `scalar` directly. The hypothetical change to the book would change
```julia-repl
julia> h = tensor!(T,2,2)
9

julia> scalar(T,4,h)
(0, Set(ParameterException{QQPolyRingElem}[(2*n1)//(q - 1) ∈ ℤ]))

julia> print_decomposition(T, h)
[...]

julia> s, e = scalar(T2,6,h); s
1

julia> e
Set{ParameterException{QQPolyRingElem}} with 2 elements:
  (2*n1)//(q - 1) ∈ ℤ
  (3*n1)//(q - 1) ∈ ℤ

```

[...]

PR #91, opened 2024-07-23
[Backport 0.2] Make compatible with Abstractalgebra 0.42

lgoettgens:
Backports #84 and #87.

This needs the creation of a release-0.2 branch and then this PR's base-branch changed to this.

This would make it possible to use #84 and #87 in proceeding with https://github.com/oscar-system/Oscar.jl/pull/3958 without needing to care about https://github.com/oscar-system/GenericCharacterTables.jl/issues/88 for the moment.


lgoettgens:
The failure on nightly is expected (https://github.com/oscar-system/Oscar.jl/issues/3888)

PR #97, opened 2024-07-23
rationalized 2F4

SoongNoonien:
Hm, maybe I have to do this again... Starting from character 29 the norm can't be computed or looks just wrong. Also `2F4.2` is still not converted.


SoongNoonien:
I've now also converted 2F4.2. After conversion the diff of both files is still essentially the same. They still differ in the exact same places, so I'm fairly certain that I didn't make any mistake. Regarding the weird norms I've mentioned earlier: After some testing with the old tables I've noticed that they where there before, so they are a separate issue. I'll open a bug with additional information later.

PR #99, opened 2024-07-24
improve show for generic characters

fingolfin:
fix a typo, and omit the paramter list if it is empty


SoongNoonien (approved):
I don't know why I did print the parameters even if they are empty...

PR #101, opened 2024-07-24
Fix and improve print_decomposition

fingolfin:
It now works again and also accepts stand alone characters
as input, not just the id of an irreducible character (type).

The printing was also slightly tweaked, and a bug fixed in the
show method for `CycloFrac` which omitted a final `Dedent()`.

This is still missing a proper test case, though... @SoongNoonien please add one

### DIFF
--- a/src/Arith.jl
+++ b/src/Arith.jl
@@ -213,6 +213,7 @@ function Base.show(io::IO, z::CycloFrac)
 		for exception in z.exceptions
 			print(io, "\n$(exception)")
 		end
+		print(io, Dedent())
 	end
 end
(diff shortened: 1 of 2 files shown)


SoongNoonien:
> This is still missing a proper test case, though... @SoongNoonien please add one

I didn't write this function, so I'm not entirely sure what the intended usecase is. I suppose a test which "decomposes" a tensor product suffices?

fingolfin:
Regarding tests, I thought about it some more and I think the best is to simply take the content of the book chapter and turn it into a section of the manual of this package. This way we can test it, and also have something to point book readers at for an always up-to-date version of the example.

PR #102, opened 2024-07-24
Print exception more nicely

### LINKED ISSUE #100: Still nicer printing of exceptions?
We currently might print something like this:
```
With exceptions:
  (2*n1)//(q - 1) ∈ ℤ
```
personally I would find this much easier to grasp:
```
With exceptions:
  2*n1 ∈ (q - 1)ℤ
```
Could we perhaps do that? Or are there complications, such as cases were it would be *worse* than what we have now?

fingolfin:
Resolves #100 -- was much easier than I expected thanks to well structured code by @SoongNoonien

### DIFF
--- a/src/Ortho.jl
+++ b/src/Ortho.jl
@@ -13,7 +13,7 @@ julia> g=genchartab("SL2.0");
 julia> classmult(g,2,2,4)
 q + 1
 With exceptions:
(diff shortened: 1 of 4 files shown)


SoongNoonien (approved):
Except for the failing doctest this looks good to me.

PR #104, opened 2024-07-24
Add OSCAR book example

fingolfin:


docs/src/book.md
```diff
@@ -0,0 +1,196 @@
...
+0
+With exceptions:
+  2*n1 ∈ (q - 1)ℤ
+```
+This scalar product is $0$
+except when $q-1$ divides $2n_1$, where $n_1$ indicates the value of the parameter $n$ for the first factor in the scalar product (i.e. the character type $\chi_4$).
+For a “generic decomposition” we need to compute all the scalar products and
+exceptions.
+```jldoctest book
+julia> print_decomposition(T, h)
```
fingolfin, line 102: Thanks to the new printing I might also just replace this by, say,
```suggestion
julia> for i in 1:8 println(i, ": ", scalar(T[i], h)) end
```
which of course produces a bit different output but it's really good enough

docs/src/book.md
```diff
@@ -0,0 +1,196 @@
...
+julia> x = param(T2, "x");  # create an additional "free" variable
+
+julia> speccharparam!(T2, 6, m, -n + (q-1)*x)  # force m = -n (mod q-1)
+```
+
+Recomputing the scalar product gives a new result
+```jldoctest book
+julia> h = tensor(T2[2],T2[2]);
+
+julia> scalar(T2[6], h)
```
fingolfin, line 176: This currently fails if I don't redefine `h` above because `h` belongs to the table `T`.

@SoongNoonien It would be nice to be able to write something like `T2(h)` to coerce `h` into the new table, *if the two table match up to the congruence.


Alternatively, perhaps it would be better to avoid the issue altogether by handling user specific congruences completely differently? I.e. instead of letting `setcongruence` produce a new table, and `speccharparam!` *modifying* a table, do it somehow else. But I am not sure how, and I think it depends a lot on practical use cases, so I'd be slow in changing this

fingolfin:
There is a bunch of other changes I made as a side product of working on this. I'lll move them gradually into separate PRs as they mature.

docs/src/book.md
SoongNoonien: Can't we remove `print_decomposition` then?

docs/src/book.md
SoongNoonien: I'm not sure if this is any good, but I'd suggest to let `setcongruence` (or better `setcongruence!`) modify the table. (We haven't decided on a behavior yet, but I'd allow setting a congruence if and only if it implies the old one. So for example allow any congruence if it previously was `nothing`, allow `(0, 2)` if it previously was `(2, 4)` but disallow `(1, 6)` if it previously was `(1, 3)`.) And I'd rename `speccharparam!` to `speccharparam` and make it return the specialized character. But this doesn't solve the problem with `specclassparam!`. I don't know the use cases for `specclassparam!`, maybe it can even be removed?

fingolfin: I agree in principle, though:
- we need to keep `specclassparam!`
- `speccharparam!` would be better called `specialize_params` or just `specialize` (the "character" bit would be implied by the type of the first argument, which would be a generic character)
- but for a user ideally there'd be a variant which takes a generic character, and a parameter exception (or a vector of parameter exceptions), and deals with it, so that the user doesn't have to come up with intermediate parameters like `x`

docs/src/book.md
fingolfin: done

docs/src/book.md
```diff
@@ -0,0 +1,179 @@
...
+
+We may now attempt to decompose this character type `h` by computing its scalar product with the irreducible character types.
+This returns a “generic” scalar product, plus a (possibly empty)
+list of parameter exceptions for which the general result may not hold.
+For example:
+```jldoctest book
+julia> scalar(T[4], h)
+0
+With exceptions:
+  2*n1 ∈ (q - 1)ℤ
```
SoongNoonien, line 79: The failing doctest here is resolved in #155.

docs/src/book.md
```diff
@@ -0,0 +1,179 @@
...
+    0
+    1
+```
+
+We may now attempt to decompose this character type `h` by computing its scalar product with the irreducible character types.
+This returns a “generic” scalar product, plus a (possibly empty)
+list of parameter exceptions for which the general result may not hold.
+For example:
+```jldoctest book
+julia> scalar(T[4], h)
```
SoongNoonien, line 76: `scalar` is just a deprecated alias for `scalar_product`

docs/src/book.md
```diff
@@ -0,0 +1,179 @@
...
+
+```
+Inspecting the list of exceptions for $\langle\chi_6,\chi_2^2\rangle$, the first
+occurs when $q-1$ divides $m+n$. To study this case we specialize $m$:
+```jldoctest book
+julia> (q, (a, b, m, n)) = params(T2);
+
+julia> x = param(T2, "x");  # create an additional "free" variable
+
+julia> speccharparam!(T2, 6, m, -n + (q-1)*x)  # force m = -n (mod q-1)
```
SoongNoonien, line 152: We don't have `speccharparam!` anymore. It got replaced by `specialize`.

docs/src/book.md
SoongNoonien: > It would be nice to be able to write something like `T2(h)` to coerce h into the new table, *if the two table match up to the congruence.

This is possible now.

SoongNoonien:
I rebased and did my best to fix the doctest fails caused by #155. I could fix all remaining failures as well but I suppose this is something to be printed with your name attached to it, so I don't want to mess around with it any further.

fingolfin:
Yes, please let me handle it from here on -- I just had to sort out a merge conflict because locally I already was further ahead in updating things :-)

docs/src/book.md
```diff
@@ -0,0 +1,178 @@
...
+```
+Inspecting the list of exceptions for $\langle\chi_6,\chi_2^2\rangle$, the first
+occurs when $q-1$ divides $m+n$. To study this case we specialize $m$ for
+character type 6:
+```jldoctest book
+julia> (q, (a, b, m, n)) = params(T2);
+
+julia> x = param(T2, "x");  # create an additional "free" variable
+
+julia> s = specialize(T2[6], m, -n + (q-1)*x)  # force m = -n (mod q-1)
```
SoongNoonien, line 152:
```suggestion
julia> s = specialize(T2[6], m, -n + (q-1)*x);  # force m = -n (mod q-1)
```

docs/src/book.md
```diff
@@ -0,0 +1,178 @@
...
+# An extended example
+
+The following is based on an example in [OSCAR](@cite),
+which in turn is based on Section 5.3 of [GHLMP96](@cite).
+To fully understand what is going on we strongly recommend to read the latter
+source concurrently with the present text.
+
+To start we load the generic character table for $\mathrm{SL}_3(q)$ with
+$q\not\equiv 1\pmod 3$.
+To learn about the origin of the table, we could enter `printinfotab(T)`.
```
SoongNoonien, line 15: Actually, we can't anymore ;-) Now it's just `info(T)`.

docs/src/book.md
```diff
@@ -0,0 +1,178 @@
...
+```jldoctest book
+julia> (q, (a, b, m, n)) = params(T2);
+
+julia> x = param(T2, "x");  # create an additional "free" variable
+
+julia> s = specialize(T2[6], m, -n + (q-1)*x);  # force m = -n (mod q-1)
+```
+
+Recomputing the scalar product now gives a new result. Note that
+we have to map the character type `h` into the new table for this work.
```
SoongNoonien, line 156:
```suggestion
we have to map the character type `h` into the new table for this to work.
```

docs/src/book.md
fingolfin, line 15:
```suggestion
To learn about the origin of the table, we could enter `info(T)`.
```

docs/src/book.md
```diff
@@ -0,0 +1,178 @@
...
+# An extended example
+
+The following is based on an example in [OSCAR](@cite),
+which in turn is based on Section 5.3 of [GHLMP96](@cite).
+To fully understand what is going on we strongly recommend to read the latter
+source concurrently with the present text.
+
+To start we load the generic character table for $\mathrm{SL}_3(q)$ with
+$q\not\equiv 1\pmod 3$.
+To learn about the origin of the table, we could enter `info(T)`.
```
fingolfin, line 15:
```suggestion
To learn about the origin of the table, we could enter `print(info(T))`.
```

docs/src/book.md
```diff
@@ -0,0 +1,178 @@
...
+  m1 - n1 ∈ (q - 1)ℤ
+  n1 ∈ (q - 1)ℤ
+<7, h> = 0
+With exceptions:
+  n1 ∈ (q - 1)ℤ
+<8, h> = 0
+With exceptions:
+  q*n1 ∈ (q^2 + q + 1)ℤ
+  n1 ∈ (q^2 + q + 1)ℤ
+  q*n1 + n1 ∈ (q^2 + q + 1)ℤ
```
fingolfin, line 111: Off-topic, but: I think this can be simplified to just
```suggestion
  n1 ∈ (q^2 + q + 1)ℤ
```
because $q$ resp. $q+1$ both are coprime with $q^2 + q + 1$ for any value of $q$. And *that* could even be determined using polynomial gcd, i.e. I think there is a chance of automating this kind of simplification: just factor the left side; then kill any factors which are provably coprime to the "modulus". 

Actually it's not simply computing a `gcd`, it is a bit more: if we start with e.g. `gcd(q, q^2+q+2)`, what I'd really want is to simplify this to `gcd(q,2)`. At that point we arrive at a case distinction: the gcd is 1 if $q$ is odd, and it is $2$ if $q$ is even.

SoongNoonien: > because q resp. q + 1 both are coprime with q 2 + q + 1 for any value of q . And that could even be determined using polynomial gcd, i.e. I think there is a chance of automating this kind of simplification: just factor the left side; then kill any factors which are provably coprime to the "modulus".

Yes, this should be easy to implement.

> Actually it's not simply computing a gcd, it is a bit more: if we start with e.g. gcd(q, q^2+q+2), what I'd really want is to simplify this to gcd(q,2). At that point we arrive at a case distinction: the gcd is 1 if q is odd, and it is 2 if q is even.

So, you are simplifying $q^2+q+2 \in q\mathbb{Z}$? As exceptions only occur in `nesum` and are always part of an exponent of a generic cyclotomic, these kind of exceptions will already be simplified. Internally the exception is $\frac{q^2+q+2}{q}=q+1+\frac{2}{q}$. And the exponents are only unique up to integers so this would be simplified to $\frac{2}{q}$.

fingolfin: I was simplifying a hypothetical `q\in (q^2+q+2)\mathbb{Z}` -- simply looked at the "problem" here and thought how one could handle it programmatically; and then thinking about ways to tweak the input to break what I just came up with -- hence arrive at this artificial example. I mainly wanted to explain why we shouldn't use `gcd(q,2)=1` (which is what Oscar gives us).

Anyway, this is for the future.

fingolfin:
We can tweak this further, but for now it is good enough to be merged.

PR #107, opened 2024-07-25
Remove setindex! method for CharTable

fingolfin:
I don't see its utility, and it is potentially quite dangerous


SoongNoonien:
I'm using this in `speccharparam!` and `specclassparam!`. Do have an alternative solution to that?

SoongNoonien:
Maybe this could be resolved like [this](https://github.com/oscar-system/GenericCharacterTables.jl/pull/104#discussion_r1691083517).

fingolfin:
Yeah as discussed there: replace `speccharparam!` by a function that applies to generic character. And change `specclassparam!` to "directly" modify the table, bypassing `setindex!`

Issue #105, opened 2024-07-25
Rename `scalar` to `scalar_product` to match OSCAR character tables


fingolfin:
The old name should stay for now but as an undocumented deprecated alias. A user reading the manual should see a single name only.

PR #108, opened 2024-07-25
Add `tensor_product` method

fingolfin:
We stills support `chi1 * chi2` for this, matching what we do for `GAPGroupClassFunction` in OSCAR, but there is a slight ambiguity there (one could also expect `*` to do a scalar product) so I think it is helpful to have an unambiguous name, too.

I chose this name to match `tensor!`, but since OSCAR has `tensor_product` and in light of #105 we should perhaps rename it to `tensor_product` ? (I'd not rename `tensor!` -- this just exists for compatibility with the book for now and will eventually be removed.

@ThomasBreuer any thoughts? Perhaps we should also support `tensor_product` for `GAPGroupClassFunction` ?

### DIFF
--- a/src/ModifyTable.jl
+++ b/src/ModifyTable.jl
@@ -1,8 +1,10 @@
-export tensor!, omega!, lincomb, lincomb!, speccharparam!, specclassparam!
+import Oscar: tensor_product
+
+export tensor_product, tensor!, omega!, lincomb, lincomb!, speccharparam!, specclassparam!
 
 # TODO deal with ParameterSubstitution, this is not done in the original implementation.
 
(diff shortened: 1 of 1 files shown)


SoongNoonien:
> I chose this name to match tensor!, but since OSCAR has tensor_product and in light of https://github.com/oscar-system/GenericCharacterTables.jl/issues/105 we should perhaps rename it to tensor_product ? (I'd not rename tensor! -- this just exists for compatibility with the book for now and will eventually be removed.

I'd say so too. Being more inline with OSCAR is always better. The behavior of `tensor`/`tensor_product` is different to the original implementation anyway, so it might as well have a different name. Renaming `tensor!` is unnecessary and maybe even confusing.

ThomasBreuer:
Yes, introducing `tensor_product` for class functions, with the same meaning as `*`, is a good idea.

PR #116, opened 2024-07-31
Add bibliography to the manual

fingolfin:
The custom citation style is copied from OSCAR and basically ensures that the exact citation keys are used that we prescribe in the .bib file.

Citation keys are instead managed via `bibtool`. To this end, the helper script `etc/format_bib` can be run from the root directory of this package to reformat `docs/src/refs.bib` in a consistent way. If a custom citation key is desired, one can add a `bibkey` field.

### DIFF
--- a/.bibtoolrsc
+++ b/.bibtoolrsc
@@ -1,7 +1,9 @@
 # This is a config file for bibtool, see
 #    http://www.gerd-neugebauer.de/software/TeX/BibTool/en/
 # To use it, invoke bibtool as follows:
-#   bibtool docs/oscar_references.bib -o docs/oscar_references.bib
+#   bibtool docs/src/refs.bib -o docs/src/refs.bib
+#
+# This file was copied from https://github.com/oscar-system/Oscar.jl version 1.1.0
 
 sort = on
 
--- a/.bibtoolrsc
+++ b/.bibtoolrsc
@@ -1,0 +1,84 @@
+# This is a config file for bibtool, see
+#    http://www.gerd-neugebauer.de/software/TeX/BibTool/en/
+# To use it, invoke bibtool as follows:
+#   bibtool docs/oscar_references.bib -o docs/oscar_references.bib
+
+sort = on
+
+# sort order for fields
+sort.order{* =
+    bibkey
+    author
+    title
+    editor
+    booktitle
+    mrnumber
+    zbl
+    journal
+    fjournal
+    series
+    volume
+    number
+    note
+    howpublished
+    address
+    organization
+    publisher
+    edition
+    pages
+    year
+    month
+    doi
+    url
+}
+
+print.align.key = 0
+print.line.length = 120
+preserve.key.case = on
+sort.cased = off
+print.use.tab = off
+
+delete.field = { hal_id }
+delete.field = { hal_version }
+delete.field = { isbn }
+delete.field = { issn }
+delete.field = { keywords }
+delete.field = { mrclass }
+delete.field = { mrreviewer }
+delete.field = { msc2010 }
+
+fmt.name.name = { }
+fmt.inter.name = { x }
+
+key.format =
+{
+    %s(bibkey)
+ #
+    %.1#n(author)
+    { %1.3n(author) }
+    { %2d(year) }
+ #
+    { %+8.1n(author) }
+    { %2d(year) }
+}
+
+# Enforce "{...}" around fields
+rewrite.rule {"^\"\([^#]*\)\"$" "{\1}"}
+rewrite.rule {"^ *\([0-9]*\) *$" "{\1}"}
+
+# Remove empty fields
+rewrite.rule {"^{ *}$"}
+
+# Unify page range separator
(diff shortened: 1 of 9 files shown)


docs/citation_style.jl
```diff
@@ -0,0 +1,38 @@
+#
+# This file is included by docs/make_work.jl to define the custom citation style `oscar_style`.
```
fingolfin, line 2: @lgoettgens I think you wrote this file -- I copied it from Oscar.jl, but there is no mention of you. I'd be happy to add a copyright notice or whatever, let me know if you have a concrete suggestion. Or we leave it as-is, whichever you prefer.

(Both packages are under the same license, so that bit at least should be fine)

docs/src/refs.bib
```diff
@@ -0,0 +1,26 @@
...
+  number        = {3},
+  note          = {Computational methods in Lie theory (Essen, 1994)},
+  pages         = {175--210},
+  year          = {1996},
+  doi           = {10.1007/BF01190329}
+}
+
+@Book{OSCAR,
+  bibkey        = {OSCAR},
+  title         = {The {C}omputer {A}lgebra {S}ystem {OSCAR}: {A}lgorithms and {E}xamples},
```
lgoettgens, line 18:
```suggestion
  title         = {The Computer Algebra System OSCAR: Algorithms and Examples},
```
DocumenterCitations does not like braces inside of fields.

docs/src/refs.bib
```diff
@@ -0,0 +1,26 @@
...
+  editor        = {Decker, Wolfram and Eder, Christian and Fieker, Claus and Horn, Max and Joswig, Michael},
+  series        = {Algorithms and {C}omputation in {M}athematics},
```
lgoettgens, line 20:
```suggestion
  series        = {Algorithms and Computation in Mathematics},
```

docs/src/refs.bib
```diff
@@ -0,0 +1,26 @@
+
+@Article{GHLMP96,
+  author        = {Geck, Meinolf and Hiss, Gerhard and Lübeck, Frank and Malle, Gunter and Pfeiffer, Götz},
+  title         = {C{HEVIE}---a system for computing and processing generic character tables},
```
lgoettgens, line 4:
```suggestion
  title         = {CHEVIE---a system for computing and processing generic character tables},
```

.bibtoolrsc
```diff
@@ -0,0 +1,84 @@
+# This is a config file for bibtool, see
+#    http://www.gerd-neugebauer.de/software/TeX/BibTool/en/
+# To use it, invoke bibtool as follows:
+#   bibtool docs/oscar_references.bib -o docs/oscar_references.bib
```
lgoettgens, line 4: path needs to be adapted

docs/citation_style.jl
lgoettgens: Hmm, what about something like the following?
```
Copyright (c) 2023 Lars Göttgens, The OSCAR Development Team
This file was copied from https://github.com/oscar-system/Oscar.jl version 1.1.0
```
I think that referencing the Oscar repository is the most important thing here.

But then we should do something similar for the other file that you copied: `.bibtoolrsc`. Either with 2021-2023 and both our names (according to its history https://github.com/oscar-system/Oscar.jl/commits/master/.bibtoolrsc), or I would be fine with just The OSCAR Development Team instead of individual names. And again add a reference to the Oscar repo, in the same style as for the other file.

Thanks a lot for thinking about such things!

docs/src/index.md
```diff
@@ -15,6 +15,23 @@ It is written in [Julia](https://julialang.org/) and depends on
...
 
+
+For the mathematical background, consider that there are many interesting
+*families* of finite groups. For example, the matrix groups
+$\mathrm{GL}_n(\mathbb{F}_q)$ or $\mathrm{SL}_n(\mathbb{F}_q)$, for $n>1$ and
+$q$ a prime power. These groups have many properties in common. It turns out
+that for a fixed rank (say $n=2$) it is possible to parametrize the conjugacy
+classes and irreducible characters of these group in terms of $q$, and to
+write this down into a so-called *generic character table*. This was first
+done by Schur for $\mathrm{SL}_2(\mathbb{F}_q)$.
```
SoongNoonien, line 26: I think in the future "Schur" should also be a reference here.

SoongNoonien (approved):
Besides the changes @lgoettgens suggested I don't see any other issues.

docs/src/index.md
fingolfin: Yeah but I'll happily leave adding all the relevant bibitems and then linking to them to you, in a future follow-up PR ;-)

docs/citation_style.jl
fingolfin: Resolved.

Though I personally really don't want to claim copyright on a simple config file like `.bibtoolrsc`.

lgoettgens: Thanks :)

PR #128, opened 2024-08-20
CI: re-enable tests against Julia 1.6

fingolfin:


SoongNoonien:
Ok, so apparently we need to adjust the `range` calls. I wanted to make them consistent anyway. Sometimes we use `start:end` and sometimes `range(start, end)`. But I don't know if `start:end` works with 1.6.

fingolfin:
`start:end` even works in 1.0

fingolfin:
Add to this one. You can push to my branch

SoongNoonien:
Hey, that did the trick already!

Issue #54, opened 2024-04-29
Properly detect redundant parameter exceptions


SoongNoonien:
As it has been seen in #51 there are places in the code where parameter exceptions are only returned after a (possibly partial) check for redundancy. It would be much better if there were a nicely documented function which decides if a parameter exception is redundant. And by redundant I mean if it isn't possibly for the exception to apply, so for example an exception `1//2+i ∈ ℤ` can safely be ignored.
But it is worth noting that `1//2+i ∈ ℤ ` shouldn't occur in the first place as we usually deal with "reduced" polynomials where reduced means without integral part.

SoongNoonien:
Also it might be worth reconsidering if a set is really the best data structure to store the exceptions. We need to compare the exceptions pairwise anyway to decide if they imply one another.

SoongNoonien:
With #155 the situation here improves. At least for all pairs of exceptions all cancellation should now be detected. But there are still exceptions which are never an integer but aren't irgnored. For example $\frac{3}{4} \in \mathbb{Z}$ or $\frac{1}{q+2} \in \mathbb{Z}$. Though I'm not sure if those can occur in the first place.

fingolfin:
Can we address this similar to what I describe in #177 ? I.e. $\frac{3}{4} \in \mathbb{Z}$ or $\frac{1}{q+2} \in \mathbb{Z}$ can be rewritten as $3 \in 4\mathbb{Z}$ or $1 \in (q+2)\mathbb{Z}$. And then trick to "factor the left side, kill any factors that are invertible modulo the right side simplifies the first to $1 \in 4\mathbb{Z}$.

Then finally, kill anything of the form `c \in n \mathbb{ZZ}` if `c` is an integer constant (here: 1) and the right hand modulus `n` is provably > 1. This is trivial in your examples. Of course it can be trickier if the modulus is e.g. $q-3$. In that case we actually do get a special case for $q=4$. And of course it can be more complicated...


That said, my instinct would be to close this for now as resolved, and open a new issue for *concrete* instances as they crop up. Most likely working towards the long-term vision outlined in #28 will yield more relevant examples

PR #131, opened 2024-08-27
added `GenericCycloFrac` using `GenericCyclo`

SoongNoonien:
This is the first follow up for #82. It introduces `GenericCycloFrac` which will replace `CycloFrac`. The code is partially copied from `Arith.jl` (which will be removed eventually). It is now independent of `ParameterException` because I like to remove this as well and I needed to somehow use `UPolyFrac` instead of `FracPoly` to model the exceptions to be compatible with `GenericCyclo`. This has the nice side effect that the `shrink` method for `GenericCycloFrac` should now be easier to understand and hopefully also easier to improve (#54).

The new code still lacks some tests.


SoongNoonien:
I'll go ahead an merge this now. The test coverage will go up again once this code actually gets used.

PR #132, opened 2024-09-03
Remove `Arith.jl`

SoongNoonien:
Finally, I've managed to remove the old `Arith.jl`. The commit history is not so nice because I faced some chicken egg problems on the way and noticed typos only some time later. The tests are still failing so I'll leave this as a draft for now.


src/CharTable.jl
```diff
@@ -3,6 +3,8 @@ export genchartab, greenfuntab
 abstract type Table{T} end
 abstract type AbstractGenericCharacter{T} end
 
+# TODO The type parameter T is not needed for CharTable and GenericCharacter
```
fingolfin, line 6: But what do you mean by that and what is stopping you from removing it? As I see it, there still is `chars::Vector{<:AbstractGenericCharacter{T}}` so it is "needed"? But perhaps `AbstractGenericCharacter` (and `Table`) can just drop it?

src/CharTable.jl
```diff
@@ -12,46 +14,44 @@ abstract type AbstractGenericCharacter{T} end
...
 	classsums::Vector{Function}  # Functions to sum a Cyclotomic over all classes in a type
 	classparamindex::Vector{Int64}  # Indices of the class parameters
 	charparamindex::Vector{Int64}  # Indices of the character parameters
-	classparams::Vector{Parameters{T}}  # Info about the parameters of each class type
-	congruence::Union{Tuple{T, T}, Nothing}  # Congruence of the main parameter q (of T). q is congruent to congruence[1] mod congruence[2].
-	modulusring::PolyRing  # Ring of polynomials of type T used in table (also ring of modulus of Cyclotomics)
-	argumentring::Generic.UniversalPolyRing{Generic.FracFieldElem{T}}  # Ring of argument of the Cyclotomics in table
+	classparams::Vector{Parameters}  # Info about the parameters of each class type
+	congruence::Union{Tuple{QQFieldElem, QQFieldElem}, Nothing}  # Congruence of the main parameter q (of T). q is congruent to congruence[1] mod congruence[2].
+	ring::GenericCycloRing  # Ring of the Cyclotomics in table  # TODO rename to parent?
```
fingolfin, line 25: This ring is not really the parent. Maybe `base_ring` would fit but that is also a bit of a stretch? `coefficient_ring`?  well, or just leave it as ring. 

So like this, including some typo fixing
```suggestion
	ring::GenericCycloRing  # parent ring of the cyclotomics in this table
```

src/CharTable.jl
fingolfin: Of course there is also `SimpleCharTable{T}` but isn't `T` essentially a constant there, too? Or can it actually take on different values?

SoongNoonien: For `SimpleCharTable{T}` it is either `PolyRingElem{QQFieldElem}` or `PolyRingElem{AbsSimpleNumFieldElem}`. For example the Green Functions of `2B2` need a square root of three.

fingolfin: Of course `SimpleCharTable` could also retain its type parameter (I see no harm in it), but that doesn't mean the abstract base type needs it? I looked at how those are used in current master and the type parameter on the abstract type was useless in all cases I saw (I may have missed some, though). E.g.:
```julia
function omega!(t::Table{T}, char::Int64) where T <: NfPoly
	if char > chartypes(t)
		throw(DomainError(char, "Character type is out of range."))
	end
	push!(t.chars, omega(t[char]))
	return length(t.chars)
end
```
could just as well be
```julia
function omega!(t::Table, char::Int64)
	if char > chartypes(t)
		throw(DomainError(char, "Character type is out of range."))
	end
	push!(t.chars, omega(t[char]))
	return length(t.chars)
end
```
(though of course we actually want to get rid of `omega!`)

SoongNoonien: > Of course SimpleCharTable could also retain its type parameter (I see no harm in it)

Ok, I'll do that.

The change of `omega!` is actually what I've done in this PR already since `T <: NfPoly` doesn't hold anymore.

SoongNoonien: >> Of course SimpleCharTable could also retain its type parameter (I see no harm in it),...

> Ok, I'll do that.

I've looked into it but I actually don't know how you mean that. The only way to get rid of the dependence on `T` I can think of is replacing all occurrences of `T` by `NfPoly` or `<:NfPoly`.

src/ModifyTable.jl
```diff
@@ -132,22 +132,22 @@ Generic character of GL2
...
+function omega(char::GenericCharacter)
 	t=parent(char)
 	charid=chartypeid(char)
-	new_char_degree=t.modulusring(1)
-	new_char_values=Vector{CycloFrac{T}}(undef, classtypes(t))
+	new_char_degree=base_ring(t.ring)(1)
+	new_char_values=Vector{GenericCyclo}(undef, classtypes(t))
 	for class in 1:classtypes(t)
-		new_char_values[class]=t.classlength[class]*char[class]//char.degree
+		new_char_values[class]=div(t.classlength[class]*char[class], char.degree)  # TODO can this really be fractional?
```
fingolfin, line 148: maybe use `divexact` (which you will have to define).

src/CharTable.jl
SoongNoonien: Ok, I got it now. It's working.

src/ModifyTable.jl
SoongNoonien: I'll do this in a follow up PR together with `omega` for `SimpleGenericCharacter` which is using `div` at the moment too.

PR #151, opened 2024-09-17
use `ngens(*)` over `length(gens(*))`

SoongNoonien:


SoongNoonien:
Oh, no. `ngens` is not defined for the universal polynomial rings.

PR #155, opened 2024-09-20
improve exception simplification

SoongNoonien:


fingolfin:
Some doctest outputs seem to have changed, but all in a way that is equivalent to what was there before.

SoongNoonien:
Yes, I'll fix them and then I think we should merge this before 0.4.0 such that the book example works again.

PR #156, opened 2024-09-21
Add nrchars method for GenericCharacter

fingolfin:


src/Show.jl
```diff
@@ -326,21 +326,29 @@ q - 1
...
+julia> g=genchartab("GL2");
+
+julia> nrchars(g[1])
+q - 1
+
+```
+"""
+function nrchars(chi::GenericCharacter)
+    t = parent(chi)
+    o=CycloSum(t.modulusring(1), t.argumentring(0))
```
SoongNoonien, line 347: I think something went wrong here. We don't have `CycloSum`, `t.modulusring` and `t.argumentring` anymore.

fingolfin: This is just a rebase of an old commit, I didn't test it afterwards. Anyway, your new PR #157 supersedes this :-)

PR #157, opened 2024-09-21
add missing methods for `AbstractGenericCharacter`

SoongNoonien:
@fingolfin I think this should be a more complete version of #156.

Closes #156

### DIFF
--- a/src/Show.jl
+++ b/src/Show.jl
@@ -1,4 +1,5 @@
 import Oscar.order
+import Oscar.AbstractAlgebra: degree
 
 export centord
 export chardeg
@@ -307,8 +308,24 @@ function chardeg(t::Table, char::Int64)
 	if char > chartypes(t)
 		throw(DomainError(char, "Character type is out of range."))
 	end
-	t[char].degree
+	return degree(t[char])
 end
+
+@doc raw"""
+    degree(char::AbstractGenericCharacter)
+
+Return the character degree of `char`.
+
+# Examples
+```jldoctest
+julia> g=genchartab("GL2");
+
+julia> degree(g[3])
+q + 1
+
+```
+"""
+degree(char::AbstractGenericCharacter) = char.degree
 
 @doc raw"""
     nrchars(t::Table, char::Int64)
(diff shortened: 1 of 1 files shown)


src/Show.jl
```diff
@@ -307,9 +307,25 @@ function chardeg(t::Table, char::Int64)
 	if char > chartypes(t)
 		throw(DomainError(char, "Character type is out of range."))
 	end
-	t[char].degree
+	return chardeg(t[char])
 end
 
+@doc raw"""
+    chardeg(char::AbstractGenericCharacter)
```
fingolfin, line 314: Can't we just call this "degree" ?

src/Show.jl
```diff
@@ -326,22 +342,36 @@ q - 1
...
-	else
-		o=t.ring(1)
-		result=t[char].sum(o//o)
-		return shrink(result)
-	end
-end
-function nrchars(t::SimpleCharTable, char::Int64)
+function nrchars(t::Table, char::Int64)
 	if char > chartypes(t)
 		throw(DomainError(char, "Character type is out of range."))
```
fingolfin, line 347: Is this even needed? Won't `t[char]` throw an error anyway? Perhaps a less nice one -- but if we want a nicer one, I think we should put it into `Base.getindex(ct::CharTable, i::Integer)`, and remove the check here.

fingolfin (approved):
Yes, much better. Though I left some comments

fingolfin:
I also think we should start moving these methods to better places. Clearly `Show.jl` is not where one would expect methods for characters to be. How about a new file `src/Chars.jl` or `src/Characters.jl` for methods applicable to characters?

And then old `nrchars` etc. which takes a table plus a charcter number should be moved to `src/Deprecations.jl`

src/Show.jl
SoongNoonien: Yes, actually I had it this way before but I thought this could be too much change at once. I'll fix it.

fingolfin:
Regarding the "moving things around": we can also just merge this now and you move later. In any case, I'll need the changes in here for the book example.

And I'd like to release 0.4 of this package this week so I can reference it in the book.

src/Show.jl
SoongNoonien: No, it is not needed. But we do have this in other functions too. See for example `printinfochar`. Shouldn't we remove it there as well?

SoongNoonien: Ok, I'll do that in a follow up PR where I also remove these checks for the class types.

Issue #166, opened 2024-09-23
Alternative to `printcharparam`


fingolfin:
We should deprecate or delete `printcharparam`. Instead there should be a function which takes a generic character and gets it parameters.

In Oscar we have `character_parameters` (CC @ThomasBreuer) but that is called on the full table, and I am not sure if those parameters are comparable to what we have. So I'd be tempted so say there simply should be a function `params` (or `parameters`?) which takes a single character and returns its `params`. Perhaps we also allow `nothing` if there are none (e.g. for simple tables). Or perhaps an empty params object should be returned, dunno, both is fine by me. Perhaps @SoongNoonien has a preference?

SoongNoonien:
I'm not a big fan of `params` as we do have this already for `CharTable` but with a slightly different meaning.

ThomasBreuer:
Concerning the attribute `character_parameters`, this refers to specializations of a generic table.
For example, the table of GL(2, q) has four types of characters (and classes), and each type has its parameter range.
```
julia> class_parameters(character_table(:GL2, 2))
3-element Vector{Tuple{Int64, Int64}}:
 (1, 0)
 (2, 0)
 (4, 1)
```
The first entry of each tuple is the type number, the second entry is the parameter value for the given type.
For q = 2, there is no character of type 3, and there is exactly one character of each of the other types.

My interpretation is that in a generic character table, each character stands for the irreducibles of its type, and the parameters of the character mean the parameter range of the type.
In this sense, the parameters of a generic table and specialized tables should be compatible, but they describe different information.

SoongNoonien:
> My interpretation is that in a generic character table, each character stands for the irreducibles of its type, and the parameters of the character mean the parameter range of the type.
In this sense, the parameters of a generic table and specialized tables should be compatible, but they describe different information.

Ok, I'd say they are different enough that `character_parameters` could be confusing.

We have replaced `printcharparam` by `params` already. So, I think we can close this issue and continue thinking about it in #180.

PR #167, opened 2024-09-23
Add params(chi::AbstractGenericCharacter)

fingolfin:
Addresses part of #166 -- it does not (yet) remove `printcharparam`.

Needed for the book example.

### DIFF
--- a/docs/src/showinfo.md
+++ b/docs/src/showinfo.md
@@ -19,6 +19,8 @@ order(t::Table)
 nrchars
 nrirrchars
 chardeg
+degree(chi::AbstractGenericCharacter)
+params(chi::GenericCharacter)
 printcharparam
 printinfochar
 ```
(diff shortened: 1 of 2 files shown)


SoongNoonien:
I'm not sure why we need this for `SimpleGenericCharacter`. In general we should reconsider where we actually need a trivial method for `SimpleGenericCharacter` or `SimpleCharTable`. But we don't provide a method of `params` for `SimpleCharTable` so I wouldn't provide a method for `SimpleGenericCharacter`.

fingolfin:
I want these methods also for `SimpleGenericCharacter` and `SimpleGenericCharacterTable` so that I can write generic code. Of course for complex algorithms it may often be best to handle the simple and "full" types separately. But for a user experimenting with these objects, it's an extra hurdle if they have to learn about the different kinds and figure out what works for one, what for the other, and what for both.

For this reason I'd also prefer if we tested more systematically with simple generic character (tables): to catch all the little things that are currently broken for those (because we never test them).

fingolfin:
In other words: to me simple vs. non-simple is in many ways an implementation detail that we should abstract away better. E.g. `params` should be added for simple tables, too, IMO.

SoongNoonien:
Ok, then I don't see anything wrong with this.

PR #176, opened 2024-09-23
Remove printval

fingolfin:


src/ModifyTable.jl
```diff
@@ -7,29 +7,23 @@ Replace the parameter `var` with `expr` in the class type `class`.
...
-julia> printval(g,class=1)
-Value of character type 1 on class type
-  1: exp(2π𝑖((2*i*k)//(q - 1)))
-Value of character type 2 on class type
-  1: q*exp(2π𝑖((2*i*k)//(q - 1)))
-Value of character type 3 on class type
-  1: (q + 1)*exp(2π𝑖((i*l + i*k)//(q - 1)))
-Value of character type 4 on class type
-  1: (q - 1)*exp(2π𝑖((i*k)//(q - 1)))
+julia> [g[i,1] for i in 1:4]
```
SoongNoonien, line 10:
```suggestion
julia> [c[1] for c in g]
```

Maybe this is a bit cleaner.

src/ModifyTable.jl
```diff
@@ -7,29 +7,23 @@ Replace the parameter `var` with `expr` in the class type `class`.
...
-julia> printval(g,class=1)
-Value of character type 1 on class type
-  1: exp(2π𝑖((6*i)//(q - 1)))
-Value of character type 2 on class type
-  1: q*exp(2π𝑖((6*i)//(q - 1)))
-Value of character type 3 on class type
-  1: (q + 1)*exp(2π𝑖((i*l + 3*i)//(q - 1)))
-Value of character type 4 on class type
-  1: (q - 1)*exp(2π𝑖((3*i)//(q - 1)))
+julia> [g[i,1] for i in 1:4]
```
SoongNoonien, line 21: Same as above.

src/GenericCharacter.jl
```diff
@@ -277,12 +277,16 @@ Return the generic character where the parameter `var` is replaced with `expr` i
 ```jldoctest
 julia> g=genchartab("GL2");
 
-julia> printval(g,char=1)
-Value of character type 1 on class type
-  1: exp(2π𝑖((2*i*k)//(q - 1)))
-  2: exp(2π𝑖((2*i*k)//(q - 1)))
-  3: exp(2π𝑖((i*k + j*k)//(q - 1)))
-  4: exp(2π𝑖((i*k)//(q - 1)))
+julia> g[1]
```
SoongNoonien, line 280: I think we should add the ability to iterate over generic characters as well. Then this would simply be:
```suggestion
julia> collect(g[1])
```
And we would only get the values.

SoongNoonien: If you think so too I can make a follow up PR addressing this.

SoongNoonien (changes requested):

src/ModifyTable.jl
fingolfin: Nice, I didn't realize we support this now.

src/GenericCharacter.jl
fingolfin: Yes that sounds like a good idea

src/ModifyTable.jl
fingolfin:
```suggestion
julia> [c[1] for c in g]
```

SoongNoonien (approved):
Ok, I'll add said PR.

PR #179, opened 2024-09-24
Remove chardeg, printcharparam, and more

fingolfin:


src/CharTable.jl
```diff
@@ -130,14 +130,13 @@ struct SimpleCharTable{T} <: Table
 	ring::PolyRing  # Ring of polynomials of type T used in table
 	information::String  # General info about the table
 	chars::Vector{<:AbstractGenericCharacter}
-	irrchartypes::Int64  # Number of irreducible character types
```
SoongNoonien, line 133: I'm a bit confused, I've already deleted this. Maybe this is based on an old branch?

fingolfin: This is definitely present in `master`. Perhaps you only deleted this in the full char table type, not in the simple one?

SoongNoonien: Oh, yes sorry.

PR #181, opened 2024-09-24
Allow iterating over characters

SoongNoonien:


src/GenericCharacter.jl
```diff
@@ -277,29 +277,21 @@ Return the generic character where the parameter `var` is replaced with `expr` i
...
+ exp(2π𝑖((2*i*k)//(q - 1)))
+ exp(2π𝑖((i*k + j*k)//(q - 1)))
+ exp(2π𝑖((i*k)//(q - 1)))
 
 julia> q,(i,j,l,k) = params(g);
 
-julia> specialize(g[1], i, q)
-Generic character of GL2
-  with parameters
-    k ∈ {1,…, q - 1}, substitutions: i = q
```
fingolfin, line 296: Actually, I think it would be good to see the printing of the substituted character (this then also serves as a test case for that).

Moreover, I am not sure if users who come to us from CHEVIE will appreciate this kind of cleverness in the documentation. They'll first have to figure out what `collect` invoked on a character even *means*, and we don't explain that anywhere (`?collect` certainly won't help either).

So while I approve of this feature, I am not sure it is wise to use it here. 

What we however *should* do (in a future PR) is expand the manual to mention that generic characters & tables implement the Julia iteration interface, and what that means, together with some examples as `jldoctest`. Something like "For example, we can use `collect` on a generic character to obtain a vector of its values:" followed by a jldoctest example.

Perhaps you can file an issue to remind us to do that?

fingolfin (approved):

PR #183, opened 2024-09-24
CI: disable tests against Julia nightly

fingolfin:
It is not helpful to always have these failing for reasons beyond our control. For background information, see here: <https://github.com/JuliaLang/julia/issues/55006>


fingolfin:
One could also turn off `depwarn` (but then we don't get warned about using deprecated stuff), or turn it off only for Julia nightly (but that requires some work). All in all I think we are OK turning it off for now.

PR #185, opened 2024-09-24
Reorganize manual

fingolfin:
Grouping things differently in a way that makes more sense to me personally.


docs/src/classes.md
```diff
@@ -0,0 +1,33 @@
...
+of class types. The results are correct for all
+possible combinations of parameters except possibly for those where the
+additionally returned exceptions apply. Those consist of multivariate
+polynomials with coefficients in a rational function field and are
+satisfied if the evaluation of this polynomial is an integer.
+
+
+```@docs
+ortho2norm
+ortho2scalar
```
fingolfin, line 31: Should we rename those to norm resp scalar_product ?

fingolfin:
@SoongNoonien any objections or concerns? I mainly moved things around and added TODO comments (I'll resolve a few later but I wanted to get the basic structure right first)

docs/src/classes.md
SoongNoonien: Yes, that's a good idea. As we don't have the old index based methods for the characters anymore we should get any conflicts.

SoongNoonien:
> @SoongNoonien any objections or concerns?

No, none. The structure of GenericCharacterTables has been deviating form the docs for quite some time now.

PR #191, opened 2024-09-24
Improve manual

### LINKED ISSUE #186: Mention new sugary bits in the docs
> Actually, I think it would be good to see the printing of the substituted character (this then also serves as a test case for that).
Moreover, I am not sure if users who come to us from CHEVIE will appreciate this kind of cleverness in the documentation. They'll first have to figure out what `collect` invoked on a character even *means*, and we don't explain that anywhere (`?collect` certainly won't help either).
So while I approve of this feature, I am not sure it is wise to use it here. 
What we however *should* do (in a future PR) is expand the manual to mention that generic characters & tables implement the Julia iteration interface, and what that means, together with some examples as `jldoctest`. Something like "For example, we can use `collect` on a generic character to obtain a vector of its values:" followed by a jldoctest example.
Perhaps you can file an issue to remind us to do that?

_Originally posted by @fingolfin in https://github.com/oscar-system/GenericCharacterTables.jl/pull/181#discussion_r1773220989_

So we should mention in the docs that one can iterate over generic characters and tables. Also the `length` and `getindex` methods are worth mentioning.

fingolfin:
Resolves #186

Locally got doctest failures, but for now I want to get this out there, as I need to switch trains now :-)

### DIFF
--- a/docs/src/characters.md
+++ b/docs/src/characters.md
@@ -1,5 +1,6 @@
 ```@meta
 CurrentModule = GenericCharacterTables
+DocTestSetup = :(using GenericCharacterTables, Oscar)
 ```
 
 # Character types
--- a/docs/src/characters.md
+++ b/docs/src/characters.md
@@ -4,20 +4,61 @@ CurrentModule = GenericCharacterTables
 
 # Character types
 
-TODO: recall what those are. Explain that "character types" are represented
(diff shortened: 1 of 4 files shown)


SoongNoonien:
I think the failures are caused by a missing `DocTestSetup = :(using GenericCharacterTables, Oscar)` at the top of the doc files.

PR #199, opened 2024-09-25
Move `info` method to `GenericCharacter.jl`

SoongNoonien:


SoongNoonien:
@fingolfin Do you have any idea why such a change could make the 1.11 test fail?

SoongNoonien:
Ok, apparently the test bots are broken. At least they refuse to test #200.

PR #206, opened 2024-09-30
Stop using `addeq!`

SoongNoonien:
This should help with https://github.com/oscar-system/Oscar.jl/pull/4163.

### DIFF
--- a/src/GenericCyclotomics.jl
+++ b/src/GenericCyclotomics.jl
@@ -2,7 +2,7 @@ using Oscar.Random: Random, SamplerTrivial, GLOBAL_RNG
 using Oscar.RandomExtensions: RandomExtensions, Make2, AbstractRNG
 
 import Oscar.AbstractAlgebra: parent_type, elem_type, base_ring, base_ring_type, parent, is_domain_type,
-	is_exact_type, canonical_unit, isequal, divexact, zero!, mul!, add!, addeq!, get_cached!, is_unit,
+	is_exact_type, canonical_unit, isequal, divexact, zero!, mul!, add!, get_cached!, is_unit,
 	characteristic, Ring, RingElem, expressify, evaluate, normal_form, divexact
 import Oscar: pretty, Indent, Dedent, terse, is_terse
 import Base: show, +, -, *, ^, ==, inv, isone, iszero, one, zero, rand, deepcopy_internal, hash, conj, promote_rule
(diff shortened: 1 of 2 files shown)


src/GenericCyclotomics.jl
```diff
@@ -365,10 +365,15 @@ end
 
 mul!(x::GenericCyclo, y::GenericCyclo, z::GenericCyclo) = y*z
 
-add!(x::GenericCyclo, y::GenericCyclo, z::GenericCyclo) = y+z
-
-function addeq!(x::GenericCyclo, y::GenericCyclo)
-	mergewith!(+, x.f, y.f)
+function add!(x::GenericCyclo, y::GenericCyclo, z::GenericCyclo)
+	if x == z
+		y,z=z,y
```
fingolfin, line 370: What if all three are equal?

SoongNoonien: In this case the swap is unnecessary and useless but not harmful.

SoongNoonien: But `==` should actually be `===` I think.

fingolfin: OK, so `mergewith` works correctly if both arguments are the same?

src/GenericCyclotomics.jl
```diff
@@ -367,10 +367,15 @@ end
...
-
-function addeq!(x::GenericCyclo, y::GenericCyclo)
-	mergewith!(+, x.f, y.f)
+function add!(x::GenericCyclo, y::GenericCyclo, z::GenericCyclo)
+	if x === z
+		y,z=z,y
+	else
+		if x !== y
+			x.f = deepcopy(y.f)
+		end
```
fingolfin, line 376:
```suggestion
	elseif x !== y
		x.f = deepcopy(y.f)
```

test/runtests.jl
```diff
@@ -60,6 +60,16 @@ test_Ring_interface(S)
...
+
+	x=deepcopy(b)
+	ab=a+b
+	@test add!(x, a, x) == ab
+	@test x == ab
+
+	add!(a, a, a)
+	@test a == 2*b
+	add!(a, c, b)
+	@test a == c+b
```
fingolfin, line 72: Ideally these tests should also verify that the other arguments are *not* changed.

fingolfin: On the long rung we should just extend the ring interface tests to cover this, as we want this kind of test for *all* implementations... I.e. `test/Rings-conformance-tests.jl:193` in AA. Not asking you to do this right now, though

fingolfin: (but see https://github.com/Nemocas/AbstractAlgebra.jl/issues/1814)

src/GenericCyclotomics.jl
SoongNoonien: Yes, I was worried too.
```
julia> a=Dict(["a" => 1, "b" => 2])
Dict{String, Int64} with 2 entries:
  "b" => 2
  "a" => 1

julia> mergewith!(+, a, a)
Dict{String, Int64} with 2 entries:
  "b" => 4
  "a" => 2


```

test/runtests.jl
SoongNoonien: Ok, I've added some extra tests for `add!` at least.

fingolfin (approved):

PR #215, opened 2024-10-07
Don't declare poly ring vars ahead of time

fingolfin:
Don't declare variables we only need later or even never.

@SoongNoonien or is there any reason why we should? If not, could you please apply the same changes to all of the tables?


SoongNoonien:
We do need them for tensor products, scalar products and class coefficients. While we could stop declaring them we will need them eventually. For a proper fix we would need to redesign the whole system with the `var_shift`s.

fingolfin:
Of course we need them but why do we have to declare them before using them? The code should be able to generate these variables on the fly, that's the whole point of universal polynomial rings, isn't it?

SoongNoonien:
> The code should be able to generate these variables on the fly, that's the whole point of universal polynomial rings, isn't it?

But the code depends on the order the variables are in. We need the universal polynomial features just for `linear_combination` where it is unknown how many variables will be needed.

PR #216, opened 2024-10-07
Fix error handling in linear_combination

fingolfin:


SoongNoonien (approved):
As we already discussed we should add a test for this. I'll try to do this in the next few days. But otherwise this looks good to me.

SoongNoonien:
Oh, we forgot the simple chars... I'll fix that.

PR #224, opened 2024-10-24
Add `Imports.jl`

SoongNoonien:


src/Imports.jl
```diff
@@ -0,0 +1,71 @@
...
+
+import Oscar:
+	class_multiplication_coefficient,
+	Dedent,
+	degree,
+	Indent,
+	norm,
+	number_of_conjugacy_classes,
+	order,
+	pretty,
```
fingolfin, line 68: Several of this are actually from AbstactAlgebra, perhaps you can move them. Or alternatively just import them all from Oscar

Issue #238, opened 2024-12-09
Benchmark scalar products for all our tables


fingolfin:
@SoongNoonien maybe you can compute timings for all our tables for ...
1. computing a single scalar product of two "random" character types,
2. computing *all* scalar products (i.e. $n^2$ if there are $n$ character types).

What are the slowest ones, and how long do they need? 

Once we have that, we can perhaps try to see how long the Maple version takes for this to get a rough idea of where we are, performance wise, at least in this particular computation.

SoongNoonien:
Ok, so here are some results for computing all scalar products:

```
GU3:   8.298026 seconds (25.19 M allocations: 3.086 GiB, 13.21% gc time, 2.87% compilation time)
PGU3.2:  10.119181 seconds (35.85 M allocations: 3.140 GiB, 12.50% gc time, 1.93% compilation time)
PGU3.n2:   3.886293 seconds (14.55 M allocations: 1.190 GiB, 14.41% gc time, 3.53% compilation time)
PSU3.2:  17.323727 seconds (64.42 M allocations: 5.543 GiB, 13.09% gc time, 1.54% compilation time)
PSU3.n2:   3.751353 seconds (14.55 M allocations: 1.190 GiB, 10.74% gc time, 3.60% compilation time)
SU3.2:  28.732406 seconds (98.44 M allocations: 8.346 GiB, 12.27% gc time, 0.91% compilation time)
SU3.n2:   4.275720 seconds (14.55 M allocations: 1.190 GiB, 15.06% gc time, 3.42% compilation time)
2B2:   1.159450 seconds (3.74 M allocations: 392.136 MiB, 12.15% gc time, 6.30% compilation time)
Sz:   1.261504 seconds (3.74 M allocations: 392.136 MiB, 16.90% gc time, 6.15% compilation time)
uni2D4.0:   0.013009 seconds (45.22 k allocations: 4.570 MiB)
uni2D4.1:   0.014082 seconds (52.61 k allocations: 5.309 MiB)
2G2:  22.078558 seconds (72.43 M allocations: 6.284 GiB, 11.78% gc time, 1.54% compilation time)
ree:  20.463347 seconds (72.26 M allocations: 6.275 GiB, 12.03% gc time, 1.06% compilation time)
3D4.0: 631.629873 seconds (2.00 G allocations: 184.452 GiB, 12.63% gc time, 0.15% compilation time)
[...]

PR #240, opened 2025-01-04
Don't recompute congruence substitutions

SoongNoonien:
As discussed earlier we shouldn't recompute the substitutions used in the simplification of `GenericCyclo`. Apparently I've made it slower and increased the allocations when computing scalar products of tables with congruence as in #238 .
Before the patch with `SL3.1`:
` 25.557065 seconds (98.68 M allocations: 8.370 GiB, 11.97% gc time)`
And after the patch with `SL3.1`:
` 36.824178 seconds (111.61 M allocations: 10.982 GiB, 9.44% gc time)`

@fingolfin Do you see the problem here?

### DIFF
--- a/src/GenericCyclotomics.jl
+++ b/src/GenericCyclotomics.jl
@@ -365,13 +365,6 @@ function (R::GenericCycloRing)(f::Dict{UPolyFrac,UPoly}; simplify::Bool=true)  #
     return GenericCyclo(f, R)
   end
 
-  # congruence preparation
-  if R.congruence !== nothing
-    q = gen(base_ring(R), 1)
-    substitute = R.congruence[2] * q + R.congruence[1]
-    substitute_inv = (q - R.congruence[1]) * 1//R.congruence[2]
-  end
-
   # reduce numerators modulo denominators
   L = NTuple{4,UPoly}[]
   for (g, c) in f
@@ -381,8 +374,8 @@ function (R::GenericCycloRing)(f::Dict{UPolyFrac,UPoly}; simplify::Bool=true)  #
       else
         gp =
           evaluate(
-            numerator(g), [1], [substitute]
-          )//evaluate(denominator(g), [1], [substitute])
+            numerator(g), [1], [R.substitute]
+          )//evaluate(denominator(g), [1], [R.substitute])
       end
       a, r = divrem(numerator(gp), denominator(gp))
       push!(L, (c, denominator(gp), r, a))
@@ -430,8 +423,8 @@ function (R::GenericCycloRing)(f::Dict{UPolyFrac,UPoly}; simplify::Bool=true)  #
       else
         gp =
           evaluate(
-            numerator(g), [1], [substitute_inv]
-          )//evaluate(denominator(g), [1], [substitute_inv])
+            numerator(g), [1], [R.substitute_inv]
+          )//evaluate(denominator(g), [1], [R.substitute_inv])
       end
       if haskey(fp, gp)
         fp[gp] += cp * c
@@ -460,13 +453,11 @@ end
 
 # Parent constructor
 
-# TODO Maybe don't require at least one variable?
 function generic_cyclotomic_ring(
   R::UPolyRing;
   congruence::Union{Tuple{ZZRingElem,ZZRingElem},Nothing}=nothing,
   cached::Bool=true,
 )
-  length(gens(R)) < 1 && error("At least one free variable is needed")
   return GenericCycloRing(R, congruence)
 end
 
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -30,6 +30,23 @@ Generic cyclotomic ring
 struct GenericCycloRing <: Ring
   base_ring::UPolyRing
   congruence::Union{Tuple{ZZRingElem,ZZRingElem},Nothing}
+  substitute::UPoly
+  substitute_inv::UPoly
+  function GenericCycloRing(
+    R::UPolyRing,
+    congruence::Union{Tuple{ZZRingElem,ZZRingElem},Nothing},
+  )
+    # TODO Maybe don't require at least one variable?
+    length(gens(R)) < 1 && error("At least one free variable is needed")
+    if congruence == nothing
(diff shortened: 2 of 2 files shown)


fingolfin:
Which Julia and OSCAR versions do you use? Did you run `@time` twice? I am using Julia 1.11.2 and Oscar 1.2.2, and am using this test code:
```julia
using GenericCharacterTables
function scalar_test(g)
   for c1 in g
       for c2 in g
           scalar_product(c1, c2)
       end
   end
end
g=generic_character_table("SL3.1")
@time scalar_test(g)
@time scalar_test(g)
```

With master:
```
julia> @time scalar_test(g)
  2.845629 seconds (30.61 M allocations: 1.738 GiB, 15.27% gc time, 26.10% compilation time)

julia> @time scalar_test(g)
  2.115675 seconds (29.62 M allocations: 1.689 GiB, 20.44% gc time)
```

With your PR:
```
julia> @time scalar_test(g)
  2.889902 seconds (30.64 M allocations: 1.786 GiB, 12.19% gc time, 26.24% compilation time)

julia> @time scalar_test(g)
  2.147462 seconds (29.64 M allocations: 1.736 GiB, 15.47% gc time)
```
So this is indeed a little bit slower and allocates a bit more, but overall much less than what you have.

SoongNoonien:
Oh, no... I still had Oscar `1.1.1` installed. Here are the results on my machine with the same versions as on yours.

Current master:
```
julia> @time scalar_test(g)
 11.932806 seconds (30.61 M allocations: 1.739 GiB, 22.33% gc time, 25.68% compilation time)

julia> @time scalar_test(g)
  7.297629 seconds (29.62 M allocations: 1.689 GiB, 21.61% gc time)
```

With the PR:
```
julia> @time scalar_test(g)
 11.660848 seconds (30.63 M allocations: 1.786 GiB, 21.91% gc time, 25.92% compilation time)

julia> @time scalar_test(g)
  7.690424 seconds (29.64 M allocations: 1.736 GiB, 21.38% gc time)
```

src/Types.jl
```diff
@@ -30,6 +30,23 @@ Generic cyclotomic ring
...
+    congruence::Union{Tuple{ZZRingElem,ZZRingElem},Nothing},
+  )
+    # TODO Maybe don't require at least one variable?
+    length(gens(R)) < 1 && error("At least one free variable is needed")
+    if congruence == nothing
+      return new(R, congruence)
+    else
+      q = gen(R, 1)
+      substitute = congruence[2] * q + congruence[1]
+      substitute_inv = (q - congruence[1]) * 1//congruence[2]
```
fingolfin, line 46: As discussed orally, computing this here in advance may be too costly. I would recommend changing the ring to a `mutable struct` (most of our ring types are anyway, to allow them to store attributes), and then leave the two fields undefined in the constructor.

Then add an accessor function (or two, depending on taste) to get their content, with suitable `isdefined(R, :substitute)` checks

fingolfin, line 46: Also, why isn't this
```suggestion
      substitute_inv = (q - congruence[1]) // congruence[2]
```

src/GenericCyclotomics.jl
```diff
@@ -381,8 +374,8 @@ function (R::GenericCycloRing)(f::Dict{UPolyFrac,UPoly}; simplify::Bool=true)  #
       else
         gp =
           evaluate(
-            numerator(g), [1], [substitute]
-          )//evaluate(denominator(g), [1], [substitute])
+            numerator(g), [1], [R.substitute]
+          )//evaluate(denominator(g), [1], [R.substitute])
```
fingolfin, line 378: If the fraction was reduced (i.e. numerator and denomiator are coprime) *before* the substitution, then it still is after, correct? But the `//` method can't know that and thus may end up trying to reduce? (I don't know if it actually does that, I just wonder).

In any case, you next need numerator and denominator anyway. So perhaps something like this could be used in this loop instead:
```julia
  den_g = denominator(g)
  num_g = numerator(g)
  if R.congruence === nothing
    den_g = evaluate(den_g, [1], [substitute])
    num_g = evaluate(num_g, [1], [substitute])
  end
  a, r = divrem(num_g, den_g)
  push!(L, (c, den_g, r, a))
```
This would arguably also make the code a little bit nicer?

src/GenericCyclotomics.jl
```diff
@@ -365,13 +365,6 @@ function (R::GenericCycloRing)(f::Dict{UPolyFrac,UPoly}; simplify::Bool=true)  #
...
-  # congruence preparation
-  if R.congruence !== nothing
-    q = gen(base_ring(R), 1)
-    substitute = R.congruence[2] * q + R.congruence[1]
-    substitute_inv = (q - R.congruence[1]) * 1//R.congruence[2]
-  end
-
   # reduce numerators modulo denominators
   L = NTuple{4,UPoly}[]
   for (g, c) in f
```
fingolfin, line 370: Style / taste question, but: I like to avoid nested control flow, esp. multi-level, to improve code clarity. So here I'd do
```suggestion
  for (g, c) in f
    is_zero(c) && continue
```
and then drop the `if !iszero(c)` and thereby reduce the indention level inside the loop

src/GenericCyclotomics.jl
SoongNoonien: > If the fraction was reduced (i.e. numerator and denomiator are coprime) before the substitution, then it still is after, correct?

No, not necessarily. This is kind of the whole point of this substitution. Take for example the fraction `(q+1)//2`. If `q` is now congruent to `1` modulo `2` then `R.substitute` is equal to `2*q+1`. After evaluation the fraction is `(2*q+2)//2` which is not reduced anymore. The idea is then that the fraction gets reduced to `q+1` which has a normal form of `0`. This detects that `exp(2*pi*i*(q+1)//2)` is equal to `1` for all `q` congruent to `1` modulo `2`.

fingolfin: ah of course, gotcha.

So please add a comment explaining just that? :-)

PR #241, opened 2025-01-12
Optimize computation of least common denominator

fingolfin:


fingolfin:
Error with Oscar 1.1: `denominator(::AbstractAlgebra.Generic.MPoly{QQFieldElem})` does not exist.

Of course that type of ring should not even be used (it should use `QQMPolyRingElem`) but that's one of the many things we fixed in Oscar 1.2 (resp. newer AA & Nemo versions). We can probably workaround it by adding a suitable such method when detecting it is needed.

SoongNoonien:
> We can probably workaround it by adding a suitable such method when detecting it is needed.

Can we detect this by checking for a specific AA version? Since we already have a check for the AA version this could be the easiest way.

fingolfin:
It may depend on the version of AA *and* Nemo, dunno. I think the easiest would be to check the version of Oscar? `if pkgversion(Oscar) < v"1.2"` ... ?

PR #244, opened 2025-01-16
Use some unsave operators

SoongNoonien:
With the test for `3D4.0` from #238 current master performs like this:
```
julia> @time scalar_test(g)
 71.137125 seconds (233.27 M allocations: 24.491 GiB, 12.05% gc time)
```
And this PR:
```
julia> @time scalar_test(g)
 66.839489 seconds (231.88 M allocations: 24.387 GiB, 11.68% gc time)
```


SoongNoonien:
Ok, after getting the tests to work again the improvement has vanished...

PR #246, opened 2025-01-22
Lazy substitute evaluation

SoongNoonien:


SoongNoonien:
I don't get it... It's still slower than master. Current master:
```
  3.241427 seconds (26.15 M allocations: 1.513 GiB, 23.60% gc time)
```

This PR:
```
  3.351800 seconds (26.20 M allocations: 1.514 GiB, 23.86% gc time)
```

This was the same test as in #240. Tomorrow I'll try to find why this is slower.

SoongNoonien:
Ok, I see. I forgot to save the substitutes...

SoongNoonien:
Ok, very good. At least some improvement:
```
  2.191820 seconds (24.12 M allocations: 1.428 GiB, 22.54% gc time)
```

PR #248, opened 2025-01-23
Initial expressify test

SoongNoonien:
I took a little closer look at the expressify system. Apart from the obvious fact that the doctests will still fail I see two other problems. The result of an ordinary scalar product is a `GenericCycloFrac` for technical reasons. This leads to scalar product results like `1//1` instead of simply `1` . Another problem is that obviously expressions like `exp(0)` are not automatically simplified to `1`. So without any manual intervention things like `0//exp(0)` will be printed.

### DIFF
--- a/src/GenericCyclotomicFractions.jl
+++ b/src/GenericCyclotomicFractions.jl
@@ -24,7 +24,10 @@ function shrink(a::GenericCycloFrac)  # TODO Move this to the constructor of Gen
 end
 
 function expressify(x::GenericCycloFrac; context = nothing)
-  return Expr(:call, ://, expressify(x.numerator, context = context), expressify(x.denominator, context = context))
+  n = expressify(x.numerator; context)
+  isone(x.denominator) && return n
+  d = expressify(x.denominator; context)
+  return Expr(:call, ://, n, d)
 end
 
 @enable_all_show_via_expressify GenericCycloFrac
--- a/src/GenericCyclotomicFractions.jl
+++ b/src/GenericCyclotomicFractions.jl
@@ -23,14 +23,33 @@ function shrink(a::GenericCycloFrac)  # TODO Move this to the constructor of Gen
   return GenericCycloFrac(new_numerator, new_denominator, a.exceptions; simplify=false)
 end
 
-function expressify(x::GenericCycloFrac; context = nothing)
-  n = expressify(x.numerator; context)
-  isone(x.denominator) && return n
-  d = expressify(x.denominator; context)
-  return Expr(:call, ://, n, d)
+function show(io::IO, x::GenericCycloFrac)
+  io = pretty(io)
+  if isone(x.denominator)
+    print(io, "$(x.numerator)")
+  else
+    if isone(length(x.numerator.f))
+      print(io, "$(x.numerator)//")
+    else
+      print(io, "($(x.numerator))//")
+    end
+    if isone(length(x.denominator.f))
+      argument, modulus = collect(x.denominator.f)[1]
+      if iszero(argument) && (is_monomial(modulus) || is_constant(modulus))
+        print(io, "$(x.denominator)")
+      else
+        print(io, "($(x.denominator))")
+      end
+    else
+      print(io, "($(x.denominator))")
+    end
+  end
+  if is_restriction(x.exceptions)
+    print(io, "\nWith exceptions:\n", Indent())
+    print(io, x.exceptions)
+    print(io, Dedent())
+  end
 end
-
-@enable_all_show_via_expressify GenericCycloFrac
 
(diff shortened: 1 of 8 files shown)


src/GenericCyclotomics.jl
```diff
@@ -211,27 +211,15 @@ function show(io::IO, R::GenericCycloRing)
...
-      print(io, "$(modulus)*exp(2π𝑖($(argument)))")
-    else
-      print(io, "($(modulus))*exp(2π𝑖($(argument)))")
-    end
-    if i < length(x.f)
-      print(io, " + ")
-    end
+    return 0
   end
+  return reduce((a, b) -> Expr(:call, :+, a, b), map(p -> Expr(:call, :*, expressify(p[2], context = context), Expr(:call, Symbol("exp"), Expr(:call, :*, Symbol("2π𝑖"), expressify(p[1], context = context)))), collect(x.f)))
```
fingolfin, line 218: while I generally enjoy higher order functions like `reduce` and `map`, I think it is rather too hard to track what's going on here. Perhaps some formatting will help; or splitting this up. 
You could also try out using `mapreduce`, but here I really think using a good old loop may be less "elegant" but probably easier to read.

Something else: instead of splitting `a + b + c` into `(a+b)+c` as you do right now (two nested `Expr`), you can use a single `Expr(:call, :+, a, b)`. Have a look at the AA and Nemo code for e.g. polynomials, you'll find stuff like this:
```julia
function expressify(@nospecialize(a::Union{PolynomialElem, NCPolyRingElem}),
   x = var(parent(a)); context = nothing)
   sum = Expr(:call, :+)
   for k in degree(a):-1:0
      c = coeff(a, k)
      if !iszero(c)
         xk = k < 1 ? 1 : k == 1 ? x : Expr(:call, :^, x, k)
         if isone(c)
            push!(sum.args, Expr(:call, :*, xk))
         else
            push!(sum.args, Expr(:call, :*, expressify(c, context = context), xk))
         end
      end
   end
   return sum
end
```

I think `collect(x.f)` could just be replaced by `x.f`.

To avoid the `exp(0)` you could just check for it yourself.

Combined, how about something like this:
```suggestion
    s = Expr(:call, :+)
    for (k,v) in x.f
      k_expr = expressify(k; context)
      v_expr = expressify(v; context)
      if is_syntactic_zero(k_expr)
        tmp = v_expr
      else
        exp_k = Expr(:call, Symbol("exp"), Expr(:call, :*, Symbol("2π𝑖"), k_expr))
        tmp = Expr(:call, :*, v_expr, exp_k)
      end
      push!(s, tmp)
    end
    return s
```

src/GenericCyclotomicFractions.jl
```diff
@@ -23,34 +23,12 @@ function shrink(a::GenericCycloFrac)  # TODO Move this to the constructor of Gen
...
-      print(io, "($(x.denominator))")
-    end
-  end
-  if is_restriction(x.exceptions)
-    print(io, "\nWith exceptions:\n", Indent())
-    print(io, x.exceptions)
-    print(io, Dedent())
-  end
+function expressify(x::GenericCycloFrac; context = nothing)
+  return Expr(:call, ://, expressify(x.numerator, context = context), expressify(x.denominator, context = context))
```
fingolfin, line 27: To avoid the `1//1` issue, you could just do this:
```suggestion
  n = expressify(x.numerator; context)
  isone(x.denominator) && return n
  d = expressify(x.denominator; context)
  return Expr(:call, ://, n, d)
```
at least if testing `isone(x.denominator)` is reasonable here... If not one can always fallback to testing purely syntactically:
```suggestion
  n = expressify(x.numerator; context)
  d = expressify(x.denominator; context)
  is_syntactic_one(d) && return n
  return Expr(:call, ://, n, d)
```

src/GenericCyclotomics.jl
SoongNoonien: > I think `collect(x.f)` could just be replaced by `x.f`.

I thought so as well, but apparently `map` doesn't support dictionaries.

fingolfin: I still think my explicit loop is superior for long-term maintainability of this code

src/GenericCyclotomicFractions.jl
SoongNoonien: The first version is sufficient here.

src/GenericCyclotomics.jl
SoongNoonien: Yes, it is. Without it it would be hard to catch `exp(0)`.

SoongNoonien:
I had to remove the expressify for `GenericCycloFrac` as they represent more than just a ring element. Namely the exceptions need to be printed as well. I think the failing nightly tests are unrelated to this PR.

PR #253, opened 2025-02-11
CI: add run with julia 1.12-nightly

lgoettgens:

PR #254, opened 2025-02-11
Fix table loading in latest julia

lgoettgens:
cc @SoongNoonien @fingolfin 

This should fix the CI jobs `(nightly, devel)` and (once https://github.com/oscar-system/GenericCharacterTables.jl/pull/253 is merged) `(1.12-nightly, devel)` and `(1.12-nightly, stable)`.

Failure of `(nightly, stable)` is expected until there is a new Oscar release that includes binaries for GAP and singular that were compiled against julia 1.13.

### DIFF
--- a/src/CharTable.jl
+++ b/src/CharTable.jl
@@ -48,9 +48,9 @@ iterate(c::AbstractGenericCharacter, state::Integer=1) =
   state > length(c) ? nothing : (c[state], state + 1)
 
 function loadtab(path::String)
-  return (@eval module $(gensym("CHAR_TABLE"))
-  include($(path))
-  end).TABLE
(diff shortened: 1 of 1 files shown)


SoongNoonien:
Thank you! But why the white space changes? The previous white spaces and newlines were generated by the code formatter.

lgoettgens:
> Thank you! But why the white space changes? The previous white spaces and newlines were generated by the code formatter.

I have no idea. But they got added by JuliaFormatter when running it on the particular file with the config in this repo.

SoongNoonien:
Ok, then I guess they are meant to be there :-)

fingolfin (approved):
Thanks!

PR #256, opened 2025-02-14
Version 0.4.3

fingolfin:
This includes a backport of PR #254


fingolfin:
I'd like to avoid a mess with the release notes again.

Perhaps best would be to first release 0.5.0 and then afterwards 0.4.3?

Or perhaps @lgoettgens has some other hints how else I can ensure only the one change compared to 0.4.2 is mentioned in the automatic release notes...

lgoettgens (approved):
If you want working tests here, you may want to backport #231 and #255 as well.
But both of these just touch test files here, so they should not do any semantic difference for downstream or book tests

PR #258, opened 2025-02-21
Cache results of `kempner_with_data` and other related computations

fingolfin:
This has only a tiny impact for our "standard example" from PR #238 but still seems useful. Tests are done against current Oscar.jl master

Before:
```
julia> @time scalar_test(g)
 18.572885 seconds (233.76 M allocations: 24.483 GiB, 10.40% gc time, 5.98% compilation time)

julia> @time scalar_test(g)
 17.501142 seconds (230.64 M allocations: 24.328 GiB, 10.84% gc time)
```

After:
```
julia> @time scalar_test(g)
 18.635003 seconds (233.66 M allocations: 24.477 GiB, 10.71% gc time, 6.01% compilation time)

julia> @time scalar_test(g)
 17.423853 seconds (230.54 M allocations: 24.322 GiB, 10.94% gc time)
```

So it just saves ~100000 allocations, but that's not nothing.

---

I think shorter & quicker micro benchmarks are useful for quick turnaround, so here is a new "standard example" we can copy & paste around:

Before:
```
julia> using GenericCharacterTables, Chairmarks

julia> g=generic_character_table("3D4.0");

julia> @b scalar_product(g[3],g[7])
3.098 ms (75572 allocs: 3.831 MiB)

julia> @b scalar_product(g[20],g[21])
77.750 ms (995574 allocs: 92.981 MiB, 13.43% gc time)
```

After:
```
julia> using GenericCharacterTables, Chairmarks

julia> g=generic_character_table("3D4.0");

julia> @b scalar_product(g[3],g[7])
3.035 ms (75572 allocs: 3.831 MiB)

julia> @b scalar_product(g[20],g[21])
77.214 ms (994886 allocs: 92.936 MiB, 13.85% gc time)
```

So no change in the first scalar product, and 690 fewer in the second (which is 0.07% ... *sigh*).

So we really need to go back to some real profiling.

### DIFF
--- a/src/GenericCyclotomics.jl
+++ b/src/GenericCyclotomics.jl
@@ -10,7 +10,16 @@ julia> kempner_with_data(12)
 
 ```
 """
-function kempner_with_data(m::Int64)
+kempner_with_data(m::Int64) = get!(() -> _kempner_with_data(m), kempner_cache, m)
+
+# FIXME/TODO: once we drop support for Oscar 1.1 we should change
+# the `Fac` in the next declaration to `Fac{Int}`.
+# Alternatively, we could remove the `factorization` return value
+# of `kempner_with_data` (a breaking change), and modify `_kempner_reduction`
+# to replace `factorization[p]` by `valuation(m, p)`
+const kempner_cache = Dict{Int,Tuple{Int, Dict{Int, Vector{Int}}, Fac}}()
+
+function _kempner_with_data(m::Int64)
   # factor m for later use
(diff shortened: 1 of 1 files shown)


fingolfin:
Made it work with Oscar 1.1 (see new code comment)

SoongNoonien:
> In the meantime: any thoughts on this PR

I'm not so sure if it is worth it. I think this makes the code less nice.

PR #269, opened 2025-02-27
Shorten `CSp6.1` loading time

SoongNoonien:
I wrote a function which converts a `GenericCyclo` to a string which represents Julia code that produces the same `GenericCyclo` but without the need for simplification. Using this I converted `CSp6.1` to a version with much faster loading time.
Before with current master:
```
julia> @time g=generic_character_table("CSp6.1");
 27.795376 seconds (137.06 M allocations: 14.305 GiB, 22.34% gc time, 0.01% compilation time)
```

With this PR:
```
julia> @time h=generic_character_table("CSp6.1");
  6.326488 seconds (10.19 M allocations: 1002.839 MiB, 19.77% gc time, 0.11% compilation time)
```

Just for reference this is the result when the simplification isn't skipped (i.e. without `simplify = false`):
```
julia> @time g=generic_character_table("CSp6.1");
  7.550312 seconds (25.35 M allocations: 2.542 GiB, 18.46% gc time, 0.05% compilation time)
```


SoongNoonien:
Just for reference, her is the function I used:
```
function cyclo_to_string(a::GenericCharacterTables.GenericCyclo, R::String="R", S::String="S")
	dict = String[]
	for (expo, fact) in a.f
		n = numerator(expo)
		d = denominator(expo)
		push!(dict, "$(R)($(n))//$(R)($(d)) => $(R)($(fact))")
	end
	if isempty(dict)
		return "$(S)()"
	else
		dict_string = join(dict, ", ")
		return "$(S)(Dict($(dict_string)), simplify=false)"
	end
end
```

PR #273, opened 2025-03-05
Replace hacky table loading by a supported interface

lgoettgens:
Thanks to @felix-roehrich for making me aware of `evalfile`.


SoongNoonien:
Nice, thank you! As far as I understand this we can even remove all the `TABLE=` at the end of the table files.

lgoettgens:
> Nice, thank you! As far as I understand this we can even remove all the `TABLE=` at the end of the table files.

Yes exactly. It's just important that the last statement of each file is the table definition

PR #274, opened 2025-03-05
Adapt to better conformance test interface

lgoettgens:
This interface was introduced in AA v0.44.5 (https://github.com/Nemocas/AbstractAlgebra.jl/issues/1936), and the old one will be removed with the next breaking AA release.

This PR adapts the uses in GCT to the new interface. However, this requires a bump to the minimum required OSCAR version. I don't know what your convention is regarding that, but you could also just keep this PR around unmerged until the old interface is removed altogether and only merge it then.


fingolfin:
We can do this but I'd like to call it a breaking change, i.e. switch to version 0.6. Also the CI tests which explicitly test against OSCAR 1.1 of course will need to be removed.

PR #276, opened 2025-03-06
Add cyclotomic factorization

SoongNoonien:
This should replace https://github.com/thofma/Hecke.jl/pull/1784. But if we merge this here we should also try to provide a method for multivariate/universal polynomials as these are also very common here.

### DIFF
--- a/src/CyclotomicFac.jl
+++ b/src/CyclotomicFac.jl
@@ -97,7 +97,7 @@ x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6
 """
 function evaluate(a::CyclotomicFac)
    r = a.unit
-   for (p, e) in Base.Iterators.flatten((a.fac, Base.Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac)))
+   for (p, e) in Iterators.flatten((a.fac, Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac)))
       r *= p^e
    end
    return r
--- a/src/CyclotomicFac.jl
+++ b/src/CyclotomicFac.jl
@@ -97,40 +97,11 @@ x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6
 """
 function evaluate(a::CyclotomicFac)
    r = a.unit
-   for (p, e) in a
+   for (p, e) in Base.Iterators.flatten((a.fac, Base.Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac)))
       r *= p^e
    end
    return r
 end
-
-function iterate(a::CyclotomicFac, state=(nothing, true))
-  if state[2]
-    if state[1] === nothing
-      fac_iter = iterate(a.fac)
-    else
-      fac_iter = iterate(a.fac, state[1])
-    end
-    if fac_iter === nothing
-      return iterate(a, (nothing, false))
-    end
-    elem, state = fac_iter
-    return (elem, (state, true))
-  else
-    if state[1] === nothing
-      cyclo_fac_iter = iterate(a.cyclo_fac)
-    else
-      cyclo_fac_iter = iterate(a.cyclo_fac, state[1])
-    end
-    if cyclo_fac_iter === nothing
-      return nothing
-    end
-    (n, expo), new_state = cyclo_fac_iter
-    pol = cyclotomic_polynomial(n, parent(a.unit))
-    return (pol => expo, (new_state, false))
-  end
-end
-
-eltype(::Type{CyclotomicFac{T}}) where {T} = eltype(Dict{T, Int})
 
 @doc raw"""
     length(a::CyclotomicFac)
--- a/src/CyclotomicFac.jl
+++ b/src/CyclotomicFac.jl
@@ -96,11 +96,15 @@ x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6
 ```
 """
 function evaluate(a::CyclotomicFac)
-   r = a.unit
-   for (p, e) in Iterators.flatten((a.fac, Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac)))
-      r *= p^e
-   end
-   return r
+  r = a.unit
+  for (p, e) in a.fac
+    r *= p^e
+  end
+  R = parent(a.unit)
+  for (n, e) in a.cyclo_fac
+    r *= cyclotomic_polynomial(n, R)^e
+  end
+  return r
 end
 
 @doc raw"""
--- a/src/CyclotomicFac.jl
+++ b/src/CyclotomicFac.jl
@@ -1,0 +1,140 @@
+@doc raw"""
+    CyclotomicFac{T <: PolyRingElem}
+
+The type representing a factorization into cyclotomic polynomials.
+"""
+struct CyclotomicFac{T <: PolyRingElem}
+  unit::T
+  fac::Dict{T, Int}
+  cyclo_fac::Dict{Int, Int}
+  function CyclotomicFac(f::Fac{T}) where T <: PolyRingElem
+    fac = Dict{T, Int}()
+    cyclo_fac = Dict{Int, Int}()
+    for (fact, expo) in f
+      b, n = is_cyclotomic_polynomial_with_data(fact)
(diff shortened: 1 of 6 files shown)


src/CyclotomicFac.jl
```diff
@@ -0,0 +1,117 @@
+@doc raw"""
+    CyclotomicFac{T <: PolyRingElem}
+The type representing a factorization into cyclotomic polynomials.
```
fingolfin, line 3:
```suggestion

The type representing a factorization into cyclotomic polynomials.
```

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,117 @@
...
+    end
+  end
+  return prod
+end
+
+@enable_all_show_via_expressify CyclotomicFac
+
+@doc raw"""
+    factor_cyclotomic(p::T) where T <: PolyRingElem -> CyclotomicFac{T}
+Return a factorization of `p` into cyclotomic polynomials. The main difference
```
fingolfin, line 52:
```suggestion

Return a factorization of `p` into cyclotomic polynomials. The main difference
```

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,117 @@
...
+julia> Qx, x = QQ["x"];
+julia> factor_cyclotomic(x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6)
+x^6*Φ_4*Φ_2^2*Φ_3*Φ_1^4
+```
+"""
+factor_cyclotomic(p::PolyRingElem) = CyclotomicFac(factor(p))
+
+@doc raw"""
+    evaluate(a::CyclotomicFac{T}) -> T
+Multiply out the factorization into a single element.
```
fingolfin, line 65:
```suggestion

Multiply out the factorization into a single element.
```

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,117 @@
...
+    pol = cyclotomic_polynomial(n, parent(a.unit))
+    return (pol => expo, (new_state, false))
+  end
+end
+
+eltype(::Type{CyclotomicFac{T}}) where {T} = eltype(Dict{T, Int})
+
+@doc raw"""
+    length(a::CyclotomicFac) -> Int
+Return the number of factors of $a$, not including the unit.
```
fingolfin, line 115:
```suggestion

Return the number of factors of $a$, not including the unit.
```

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,117 @@
...
+
+@enable_all_show_via_expressify CyclotomicFac
+
+@doc raw"""
+    factor_cyclotomic(p::T) where T <: PolyRingElem -> CyclotomicFac{T}
+Return a factorization of `p` into cyclotomic polynomials. The main difference
+to `factor` is the way the result gets printed.
+# Examples
+```jldoctest
+julia> Qx, x = QQ["x"];
```
fingolfin, line 56:
```suggestion
julia> Qx, x = QQ["x"];

```

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,117 @@
...
+@doc raw"""
+    evaluate(a::CyclotomicFac{T}) -> T
+Multiply out the factorization into a single element.
+# Examples
+```jldoctest
+julia> Qx, x = QQ["x"];
+julia> f = x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6;
+julia> fac = factor_cyclotomic(f)
+x^6*Φ_4*Φ_2^2*Φ_3*Φ_1^4
+julia> evaluate(fac)
```
fingolfin, line 72:
```suggestion
julia> Qx, x = QQ["x"];

julia> f = x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6;

julia> fac = factor_cyclotomic(f)
x^6*Φ_4*Φ_2^2*Φ_3*Φ_1^4

julia> evaluate(fac)
```

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,125 @@
...
+"""
+struct CyclotomicFac{T <: PolyRingElem}
+  unit::T
+  fac::Dict{T, Int}
+  cyclo_fac::Dict{Int, Int}
+  function CyclotomicFac(f::Fac{T}) where T <: PolyRingElem
+    fac = Dict{T, Int}()
+    cyclo_fac = Dict{Int, Int}()
+    for (fact, expo) in f
+      b, n = is_cyclotomic_polynomial_with_data(fact)
```
lgoettgens, line 14: This function was only added in https://github.com/thofma/Hecke.jl/pull/1779, so this PR should also set a compat entry for Hecke to 0.35.13 (which will also automatically restrict Oscar to at least 1.3)

fingolfin: Indeed, @SoongNoonien please do that

SoongNoonien: Ok, we are now restricted to Oscar 1.3 anyway.

lgoettgens: Still, Oscar 1.3.0 is compatible with Hecke 0.35.10, but you use a feature from Hecke 0.35.13. This needs a compat entry for Hecke. See e.g. https://github.com/oscar-system/Oscar.jl/pull/4592/files#diff-72ed386c2a0cd1d23c0968297e70023ed98c22490d146dd89fc91f48369bad4d for an example of how to add a compat entry for an indirect dependency

SoongNoonien: Ok, I see. Thanks!

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,140 @@
...
+      cyclo_fac_iter = iterate(a.cyclo_fac, state[1])
+    end
+    if cyclo_fac_iter === nothing
+      return nothing
+    end
+    (n, expo), new_state = cyclo_fac_iter
+    pol = cyclotomic_polynomial(n, parent(a.unit))
+    return (pol => expo, (new_state, false))
+  end
+end
```
fingolfin, line 131: Alternatively I think you could also do this (after adding `import Iterators` somewhere)
```suggestion
function iterate(a::CyclotomicFac)
  return Iterators.flatten((a.fac, a.cyclo_fac))
end
```

SoongNoonien: This would only return an iterator but not actually iterate over it. We could store this iterator in the `CyclotomicFac` struct and then wrap it with `iterate(a::CyclotomicFac)`. But we only need the iteration in `evaluate` so I removed `iterate(a::CyclotomicFac)`.

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,111 @@
...
+julia> fac = factor_cyclotomic(f)
+x^6*Φ_4*Φ_2^2*Φ_3*Φ_1^4
+
+julia> evaluate(fac)
+x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6
+```
+"""
+function evaluate(a::CyclotomicFac)
+   r = a.unit
+   for (p, e) in Base.Iterators.flatten((a.fac, Base.Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac)))
```
lgoettgens, line 100:
```suggestion
   for (p, e) in Iterators.flatten((a.fac, Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac)))
```

SoongNoonien: Ah, yes. This makes the line even shorter ;-) Thank you!

src/CyclotomicFac.jl
```diff
@@ -0,0 +1,111 @@
...
+
+julia> evaluate(fac)
+x^16 - x^15 - x^14 + 2*x^11 - x^8 - x^7 + x^6
+```
+"""
+function evaluate(a::CyclotomicFac)
+   r = a.unit
+   for (p, e) in Iterators.flatten((a.fac, Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac)))
+      r *= p^e
+   end
```
fingolfin, line 102: How about?
```suggestion
   for (p, e) in a.fac
      r *= p^e
   end

   for (p, e) in Iterators.map(x -> cyclotomic_polynomial(first(x), parent(a.unit)) => last(x), a.cyclo_fac))
      r *= p^e
   end
```
and then simplify the second loop?

fingolfin (approved):
Maybe simplify as indicated, but anyway, good to go

PR #278, opened 2025-03-12
Improve printing and construction of `GenericCyclo`

### LINKED ISSUE #261: Printing and construction of `GenericCyclo`
Currently the `GenericCyclo` elements are printed using special unicode characters like `π` and `𝑖`. This has two main disadvantages. Namely that unicode characters are hard to type and both of these symbols are printed over and over again without adding any information.

Things like this:
```q*exp(2π𝑖*2//(q - 1))```

Could also be printed like this, which is similar to how GAP prints roots of unity:

```q*E(q - 1)^2```

This is more compact and can easily be typed. To be of actual use, this `E` should have a meaning in GenericCharacterTables in such a way that entering the above will result in an actual `GenericCyclo` which has the same value. But this `E` has to be dependent on the `GenericCycloRing` because things like `E(3)` are in no relation to any `GenericCycloRing`.

### PR DESCRIPTION
**Pull request by SoongNoonien**
This should fix #261. The doctests will need a complete refactor.

### DIFF
--- a/src/GenericCyclotomics.jl
+++ b/src/GenericCyclotomics.jl
@@ -134,6 +134,8 @@ base_ring(R::GenericCycloRing) = R.base_ring::base_ring_type(typeof(R))
 
 parent(f::GenericCyclo) = f.parent
 
+parent(E::GenericCycloRingGen) = E.parent
+
 is_domain_type(::Type{GenericCyclo}) = false
 
 is_exact_type(::Type{GenericCyclo}) = true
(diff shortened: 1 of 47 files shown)

### CODE REVIEW

File: src/GenericCyclotomics.jl [hunk 1]
```diff
@@ -457,10 +471,19 @@ end
 function generic_cyclotomic_ring(
   R::UPolyRing;
   congruence::Union{Tuple{ZZRingElem,ZZRingElem},Nothing}=nothing,
+  symbol::Symbol=:E,
   cached::Bool=true,
```
fingolfin, line 475:
```suggestion
function generic_cyclotomic_ring(
  R::UPolyRing,
  symbol::Symbol=:E;
  congruence::Union{Tuple{ZZRingElem,ZZRingElem},Nothing}=nothing,
  cached::Bool=true,
```

fingolfin:
Is this now ready for review? it is still marked as draft...

SoongNoonien:
Ah, yes.

PR #279, opened 2025-03-21
Improve `^` for `GenericCyclo`

SoongNoonien:
As discussed earlier I've adjusted `^` for `GenericCyclo` to allow rational exponents. While thinking about tests to implement I've noticed that the original implementation was flawed. Things like `(2*E(3))^(1//2)` would silently return `2*E(6)`. Also `E(2)^(1//3)` returned `-1` while `E(3)^(1//2)` returned `E(6)` as I would have expected. With the adjustments needed for these things to fix, some return values might now be a bit unexpected. For example `E(1)^(1//3)` now returns `E(3)`. At least to me that is not a problem, I think I even like it.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -82,6 +82,10 @@ end
   a = S(1; exponent=(2 * i * j) * 1//R((q - 1)))
   @test isone(a^(q-1))
   @test a == a^q
+
+  @test E(1)^(1//2) == E(2)
+  @test E(2)^(1//3) == E(6)
+  @test E(3)^(1//2) == E(6)
 end
 
 @testset "Shifts" begin
--- a/src/GenericCyclotomics.jl
+++ b/src/GenericCyclotomics.jl
@@ -286,12 +286,20 @@ end
 
 *(x::GenericCyclo, y::UPoly) = y * x
 
-function ^(x::GenericCyclo, y::UPoly)
-  if isone(length(x.f))
-    expo, c = collect(x.f)[1]
-    return parent(x)(Dict(y*expo => c))
+function ^(x::GenericCyclo, y::Union{UPoly, Rational})
+  isone(length(x.f)) || error("cyclotomic has to many summands")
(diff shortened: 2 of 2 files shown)


test/runtests.jl
```diff
@@ -82,6 +82,10 @@ end
   a = S(1; exponent=(2 * i * j) * 1//R((q - 1)))
   @test isone(a^(q-1))
   @test a == a^q
+
+  @test E(1)^(1//2) == E(2)
+  @test E(2)^(1//3) == E(6)
+  @test E(3)^(1//2) == E(6)
```
fingolfin, line 88: I am conflicted about this. Last time we talked about it, didn't we agree that `E(n)^(p//q)` should be allowed (it throws a `MethodError` right now), but that the values returned by `E(n)^R(p//q)` are otherwise fine. 

In particular I dislike these values. I'd really expect this:
```suggestion
  @test E(1)^(1//2) == E(1) ==  1
  @test E(2)^(1//3) == E(2) == -1
  @test E(3)^(1//2) == E(6)
```

However, I guess a strong argument for your choices is consistency with `QQAbField`:
```
julia> root(E(1), 2)
-1

julia> root(E(2), 3)
ζ(6)

julia> root(E(3), 2)
ζ(6)

julia> E(6)
-ζ(3)        # huh???
```

Then again, `E(3)^(1//2)` does not work there. And `root(x,n::Int)` is documented as 
> Return the principal `n`-th root of `x`. We require `x \geq 0`.


Aha but `QQBar` closes the loop:
```
julia> L = algebraic_closure(QQ)
Algebraic closure of rational field

julia> L(E(3))
Root -0.500000 + 0.866025*im of x^2 + x + 1

julia> L(E(3))^(1//2)
Root 0.500000 + 0.866025*im of x^2 - x + 1

julia> L(root(E(3),2))
Root 0.500000 + 0.866025*im of x^2 - x + 1
```


So I guess we need this patch for consistency.

PR #281, opened 2025-03-21
Fix `SL2.1`

### LINKED ISSUE #275: `SL2.1` is broken
@frankluebeck spotted a flaw in the handling of generic square roots of `q`. Right now we simply substitute all occurrences of `q` in `SL2.1` by `q0^2` to be able to use `q0` as a square root of `q`. But all simplifications now assume that `q0` is an integer even worse an integer congruent to one modulo two. Which results in falsely "simplified" entries in the table.

I think we need a better way to represent the square root of `q`. If we can't find a better way we could alternatively try to improve the congruence substitution in the simplification process. Right now we replace `q0` by `2*q0+1` before the simplification and then afterwards `q0` is replaced by `1//2*(q0-1)`. If we could find a way to replace all `q0^2` by `2*q0+1` and then `q0` by `1//2*(q0^2-1)` the simplification would work correctly again. Note that at least mathematically this is no problem since `q0` will only appear with even exponents inside the exponents of the generic cyclotomic numbers.

SoongNoonien:
This should fix #275. I'm not sure if `power` is a good name though.

### DIFF
--- a/data/Tables/A1/SL2.1.jl
+++ b/data/Tables/A1/SL2.1.jl
@@ -5,7 +5,7 @@ using ..GenericCharacterTables.Oscar
 R = universal_polynomial_ring(QQ; cached=false)
 q0 = gen(R, "q0")
 q = q0^2
-S, E = generic_cyclotomic_ring(R; congruence=ZZ.((1, 2)))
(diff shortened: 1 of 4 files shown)


SoongNoonien:
Ok, I'll go ahead and merge this now and adjust the `generic_cyclotomic_ring` constructor later accordingly.

PR #282, opened 2025-03-26
Add Greenfunctions table for `E8c2`

SoongNoonien:
This file is based on a test file from @frankluebeck. It should be correct now. But not very elegant.


fingolfin:
As you just pointed out to me orally, it would be good to have an info string in this. Perhaps @frankluebeck can provide us with a few lines about the origin of this table. Also, what does "c2" mean in "E8c2" ?

frankluebeck:
This file is really just intended as a temporary larger example for testing.

Of course, the information entry could be extended a bit, say:

```
information = "generalized Green functions for E_8(q) with q = 2 mod 30, computed by Frank Lübeck"
```

I cannot make further comments because after a `git pull` the package does no longer compile (I get error messages saying that `ConformanceTests` is not defined). So, I cannot test the current version.

When I understand the code in `E8c2.jl` correctly, this version of the file stores all entries as polynomials in the square root of q. But almost all entries are polynomials in q. So, almost all polynomials are just blown up, need double space and the arithmetic is less efficient. 

Isn't there a way to compute in a quadratic extension of K[q] with most entries in this base ring?
Does anyone think about having Puiseux polynomials in Oscar?

SoongNoonien:
> This file is really just intended as a temporary larger example for testing.

So it isn't of any actual interest?

> I cannot make further comments because after a `git pull` the package does no longer compile (I get error messages saying that `ConformanceTests` is not defined). So, I cannot test the current version.

It will most certainly compile after a `]up` you need at least OSCAR 1.3.

fingolfin:
> So, almost all polynomials are just blown up, need double space and the arithmetic is less efficient.

Multivariate polynomials are stored sparsely and so only one int entry in each exponent vector gets doubled.

SoongNoonien:
> Multivariate polynomials are stored sparsely and so only one int entry in each exponent vector gets doubled.

Unfortunately, we are dealing with univariate polynomials in this case.

SoongNoonien:
I've updated the comment accordingly.

Issue #226, opened 2024-11-07
Add new page(s) to documentation listing all our tables


fingolfin:
Two ideas for using the bib data when we have it:
1. generate a page / pages in the manual with a list of all table, and also for each table
2. for each paper, list which tables resulted from it

In an ideal world these connections (which form a graph) would be computable by querying our data (that means each table object would need to contain a vector/list of the bib items it references)

_Originally posted by @fingolfin in https://github.com/oscar-system/GenericCharacterTables.jl/issues/115#issuecomment-2362990637_

PR #289, opened 2025-05-14
Add overview of all tables

fingolfin:
This is a first step towards issue #226.

Many more references need to be updated in the tables files, perhaps @SoongNoonien can take care of that.

Further tweaks to the content are also needed, but this seems like a good start...

### DIFF
--- a/data/Greenfunctions/2A2/GU3.jl
+++ b/data/Greenfunctions/2A2/GU3.jl
@@ -27,17 +27,9 @@ information = raw"""- Information about the tables of Green functions for GU_3(q
 
 - By a theorem of Hotta, Springer and Kawanaka we can get the Green
   functions of the unitary group GU_3(q^2) from those of GL_3(q) 
-  by substituting q by -q. This is proved in:
-  {\sc R.~Hotta and T.~A.~Springer}, A specialisation theorem for
-  certain Weyl group representations, {\em Invent. Math.} 
-  {\bf 41} (1977), 113--127.
-  {\sc N.~Kawanaka}, Generalized Gelfand--Graev characters and
-  Ennola duality, {\em Adv. Stud. Pure Math.} {\bf 6} (1985), 175--206.
+  by substituting q by -q. This is proved in [HS77](@cite) and [Kaw85](@cite).
 
-- See also:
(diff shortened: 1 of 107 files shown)


data/Greenfunctions/2A2/GU3.jl
```diff
@@ -27,17 +27,9 @@ information = raw"""- Information about the tables of Green functions for GU_3(q
...
-  Ennola duality, {\em Adv. Stud. Pure Math.} {\bf 6} (1985), 175--206.
+  by substituting q by -q. This is proved in [HS77](@cite) and [Kaw85](@cite).
 
-- See also:
-  {\sc F.~Digne and J.~Michel}, Foncteurs de Lusztig et caract\``{e}res
-  des groupes lin\'{e}aires et unitaires sur un corps fini, 
-  {\em J. of Alg.} {\bf 107} (1987), 217--255.
+- See also: [DM87*1](@cite).
 
 - For the computation of the Green functions for GL_n(q) see for example:
```
fingolfin, line 34:
```suggestion
- For the computation of the Green functions for $GL_n(q)$ see for example:
```

data/Greenfunctions/2A2/GU3.jl
```diff
@@ -27,17 +27,9 @@ information = raw"""- Information about the tables of Green functions for GU_3(q
...
   > GreenFunTab(GL2);
```
fingolfin, line 35: This should be turned into a proper Markdown code example

data/Greenfunctions/2A3/GU4.jl
```diff
@@ -54,17 +54,9 @@ information = raw"""- Information about the tables of Green functions for GU_4(q
 
 - By a theorem of Hotta, Springer and Kawanaka we can get the Green
   functions of the unitary group GU_4(q^2) from those of GL_4(q)
```
fingolfin, line 56:
```suggestion
  functions of the unitary group $\mathrm{GU}_4(q^2)$ from those of $\mathrm{GL}_4(q)$ 
```

fingolfin: ... and so on, @SoongNoonien please apply similar changes to the other files

fingolfin:
This is good enough for now. We can and should improve it further but having it online now will be helpful already.

PR #294, opened 2025-05-23
Add table name synonyms, nicer table lists rendering (sort table types better, nicer table type names)

### LINKED ISSUE #226: Add new page(s) to documentation listing all our tables

### LINKED ISSUE #290: Replace symbolic links for some of our tables with a list of aliases
... and then also show the alias names in our "List of tables" in the manual

fingolfin:
Resolves #290

I think we can also close #226 now -- that issue also lists the idea of reporting for each paper were it is used, but I am not sure how useful that is. Let's wait if someone requests that.


fingolfin:
Docu preview is at https://oscar-system.github.io/GenericCharacterTables.jl/previews/PR294/ so that worked. I am not sure what we have to do to get the preview link in the list of checks shown by GitHub here. Perhaps @lgoettgens or @benlorenz know?

PR #298, opened 2025-05-28
Specialization to ordinary table

SoongNoonien:
This is a first and very crude hack to get an ordinary character table from a generic one. Apparently I'm doing something wrong or the parameter specifications of `GL2` are incorrect. This is what I get for $\mathrm{GL}(2,3)$:
```
julia> g=generic_character_table("GL2")
Generic character table GL2
  of order q^4 - q^3 - q^2 + q
  with 4 irreducible character types
  with 4 class types
  with parameters (i, j, l, k)

julia> get_ordinary(g, 3)
10×10 transpose(::Matrix{GenericCharacterTables.GenericCyclo}) with eltype GenericCharacterTables.GenericCyclo:
 1   1  1   1   -1  -1              1   1   -1              1
 1   1  1   1   1   1               1   1   1               1
 3   3  0   0   -1  1               -1  -1  1               -1
 3   3  0   0   1   -1              -1  -1  -1              -1
 -4  4  -1  1   0   0               0   0   0               0
 -2  2  1   -1  0   -E(8) - E(8)^3  0   2   E(8) + E(8)^3   -2
 2   2  -1  -1  0   0               2   -2  0               -2
 2   2  -1  -1  0   2               -2  -2  2               -2
 -2  2  1   -1  0   E(8) + E(8)^3   0   2   -E(8) - E(8)^3  -2
 2   2  -1  -1  0   -2              -2  -2  -2              -2

```

There are two additional rows and columns as can be verified [here](https://ncatlab.org/nlab/show/character+table+of+GL%282%2C3%29).

### DIFF
--- a/src/Exports.jl
+++ b/src/Exports.jl
@@ -9,6 +9,7 @@ export evaluate
 export factor_cyclotomic
 export generic_character_table
 export generic_cyclotomic_ring
+export get_ordinary
 export green_function_table
 export info
 export kempner
--- a/src/Exports.jl
+++ b/src/Exports.jl
@@ -8,6 +8,7 @@ export evaluate
 export factor_cyclotomic
 export generic_character_table
 export generic_cyclotomic_ring
+export get_ordinary
 export green_function_table
 export info
 export kempner
--- a/src/GenericCharacterTables.jl
+++ b/src/GenericCharacterTables.jl
@@ -37,5 +37,6 @@ include("Congruence.jl")
 include("Exports.jl")
 include("Deprecations.jl")
 include("CyclotomicFac.jl")
+include("GetOrdinary.jl")
 
 end
--- a/src/GetOrdinary.jl
+++ b/src/GetOrdinary.jl
@@ -13,17 +13,16 @@ function get_ordinary(char::GenericCharacter, val)
   params = parameters(char)
   exceptions = params.exceptions
   ranges = UnitRange{Int64}[]
-  vars = Int64[]
+  vars = Int64[1]
   for param in params
     push!(vars, var_index(param.var))
     upper = evaluate(param.modulus, [q], [val])
     push!(ranges, 1:numerator(constant_coefficient(upper)))
   end
-  spec = specialize(char, q, val)
   for vals in Iterators.product(ranges...)
-    vals = collect(vals)
+    vals = collect((val, vals...))
     if !exceptions_apply(exceptions, vars, vals)
-      ordinary_char = evaluate.(spec.values, Ref(vars), Ref(vals))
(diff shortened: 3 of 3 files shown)


fingolfin:
You can also check the table by asking... OSCAR ;-)
```
julia> character_table(GL(2,3))
Character table of GL(2,3)

  2  4  1  4  1  3            3            3  2
  3  1  1  1  1  .            .            .  .

    1a 6a 2a 3a 4a           8a           8b 2b

X_1  1  1  1  1  1            1            1  1
X_2  1  1  1  1  1           -1           -1 -1
X_3  2 -1  2 -1  2            .            .  .
X_4  2  1 -2 -1  . -z_8^3 - z_8  z_8^3 + z_8  .
X_5  2  1 -2 -1  .  z_8^3 + z_8 -z_8^3 - z_8  .
X_6  3  .  3  . -1            1            1 -1
X_7  3  .  3  . -1           -1           -1  1
X_8  4 -1 -4  1  .            .            .  .
```
Also note that the first column should have only positive integer entries (those should be the degrees of the corresponding representations). So indeed something is wrong

SoongNoonien:
> You can also check the table by asking... OSCAR ;-)

Of course I should have thought of that...

> Also note that the first column should have only positive integer entries (those should be the degrees of the corresponding representations). So indeed something is wrong

Yes, but that's just a convention on the order of the conjugacy classes. My first goal is to obtain a table which is correct up to permutation of columns and rows. I'm not even sure if we can say beforehand which combination of class parameters corresponds to the trivial class.

SoongNoonien:
Ok, I found the problem. Now the table looks like this:
```
julia> get_ordinary(g,3)
8×8 transpose(::Matrix{GenericCharacterTables.GenericCyclo}) with eltype GenericCharacterTables.GenericCyclo:
 1   1  1   1   -1  -1              1   -1
 1   1  1   1   1   1               1   1
 3   3  0   0   -1  1               -1  1
 3   3  0   0   1   -1              -1  -1
 -4  4  -1  1   0   0               0   0
 -2  2  1   -1  0   -E(8) - E(8)^3  0   E(8) + E(8)^3
 2   2  -1  -1  0   0               2   0
 -2  2  1   -1  0   E(8) + E(8)^3   0   -E(8) - E(8)^3

```

frankluebeck:
Yes, this way it works "in principle", provided the character parameters and class parameters are correctly described.

But it quickly becomes non-practical because  for larger tables and/or parameters you produce far too many rows and columns which are then collapsed afterwords.

With newly created tables I will later provide additional information on the parameter sets, such that each class and each irreducible character is described by a unique parameter specialization.

PR #319, opened 2026-05-11
Add missing exceptions to SL3.n1

SoongNoonien:
While working through the [extended example](https://oscar-system.github.io/GenericCharacterTables.jl/dev/book/) from the docs, one might wonder if the given decomposition is actually correct. Besides checking the degree it is also intersting to see if the values on all classes coincide with the tensor product. On the conjugacy classes of type 6 the tensor product has value $4$. The partial sum consisting of the first three irreducible character types has the value $7$. So the two "generic" sums consisting of the character types 6 and 7 should have a value of $-3$. The sum of character type 7 has value $0$. Checking the sum of character type 6 we get:
```
GenericCharacterTables.nesum(s[6], n, 1, q-2)
-6
With exceptions:
  a - b ∈ (q - 1)ℤ
  a + 2*b ∈ (q - 1)ℤ
  2*a + b ∈ (q - 1)ℤ
```
This looks reasonable if we remember the factor of $\frac{1}{2}$. The first parameter exception can never apply since it is explicitly stated in the description of the conjugacy class type 6:
```
julia> T[:,6]
Generic conjugacy class of SL3.n1
  with parameters
    a ∈ {1,…, q - 1}, b ∈ {1,…, q - 1} except a - b ∈ (q - 1)ℤ
  of order q^6 + 2*q^5 + 2*q^4 + q^3
  with values
    1: 1
    2: 2
    3: 1
    4: E(q - 1)^(a*n) + E(q - 1)^(b*n) + E(q - 1)^(-a*n - b*n)
    5: E(q - 1)^(a*n) + E(q - 1)^(b*n) + E(q - 1)^(-a*n - b*n)
    6: E(q - 1)^(-a*m - b*m + b*n) + E(q - 1)^(a*m - a*n - b*n) + E(q - 1)^(-a*n + b*m - b*n) + E(q - 1)^(-a*m + a*n - b*m) + E(q - 1)^(a*m + b*
    n) + E(q - 1)^(a*n + b*m)
    7: 0
    8: 0
```
But what about the other two exceptions? Unfortunately, we get a different result there:
```
julia> GenericCharacterTables.nesum(evaluate(s[6], [a], [-2*b]), n, 1, q-2)
2*q - 8
With exceptions:
  3*b ∈ (q - 1)ℤ
```
Checking the original [paper](https://doi.org/10.4153/CJM-1973-049-7) which introduced the table one can see that the class parameter description is missing a third parameter. In the paper the three parameters are called $k$, $l$, $m$. It is easy to see from the relations that $m \equiv_{q-1} -k-l$, hence the third parameter can be dropped. But what can't be dropped is the relation $k < l < m$. With the other relations this is equivalent to $k \not\equiv_{q-1} l$, $l \not\equiv_{q-1} m$ and $m \not\equiv_{q-1} k$. Without the third parameter this translates to the following forbidden relations (i.e. exceptions): $k - l \equiv_{q-1} 0$, $k + 2l \equiv_{q-1} 0$ and $l +2k \equiv_{q-1} 0$. Only the first was present in the table and the other two directly correspond to the exceptions from above. So, in fact the weird value of `2*q - 8` we computed above is invalid and does not correspond to a legal combination of parameters.

Similarly, the character type 6 was missing two exceptions as well.

### DIFF
--- a/data/Tables/A2/SL3.n1.jl
+++ b/data/Tables/A2/SL3.n1.jl
@@ -233,7 +233,7 @@ classparams = [
   Parameters(Parameter[]),
   Parameters([Parameter(a, q - 1)], [((a) * 1//(q - 1))]),
   Parameters([Parameter(a, q - 1)], [((a) * 1//(q - 1))]),
-  Parameters([Parameter(a, q - 1), Parameter(b, q - 1)], [((a - b) * 1//(q - 1))]),
+  Parameters([Parameter(a, q - 1), Parameter(b, q - 1)], [((a - b) * 1//(q - 1)), ((a + 2*b) * 1//(q - 1)), ((2*a + b) * 1//(q - 1))]),
   Parameters([Parameter(a, q^2 - 1)], [((a) * 1//(q + 1))]),
   Parameters([Parameter(a, q^2 + q + 1)], [((a) * 1//(q^2 + q + 1))]),
 ]
@@ -244,7 +244,7 @@ charparams = [
   Parameters(Parameter[]),
   Parameters([Parameter(n, q - 1)], [((n) * 1//(q - 1))]),
   Parameters([Parameter(n, q - 1)], [((n) * 1//(q - 1))]),
-  Parameters([Parameter(n, q - 1), Parameter(m, q - 1)], [((n - m) * 1//(q - 1))]),
+  Parameters([Parameter(n, q - 1), Parameter(m, q - 1)], [((n - m) * 1//(q - 1)), ((n + 2*m) * 1//(q - 1)), ((2*n + m) * 1//(q - 1))]),
   Parameters([Parameter(n, q^2 - 1)], [((n) * 1//(q + 1))]),
   Parameters([Parameter(n, q^2 + q + 1)], [((n) * 1//(q^2 + q + 1))]),
 ]
--- a/data/Tables/A2/SL3.n1.jl
+++ b/data/Tables/A2/SL3.n1.jl
@@ -244,7 +244,7 @@ charparams = [
   Parameters(Parameter[]),
   Parameters([Parameter(n, q - 1)], [((n) * 1//(q - 1))]),
   Parameters([Parameter(n, q - 1)], [((n) * 1//(q - 1))]),
-  Parameters([Parameter(n, q - 1), Parameter(m, q - 1)], [((n - m) * 1//(q - 1)), ((n + 2*m) * 1//(q - 1)), ((2*n + m) * 1//(q - 1))]),
+  Parameters([Parameter(n, q - 1), Parameter(m, q - 1)], [((n - m) * 1//(q - 1)), ((n) * 1//(q - 1)), ((m) * 1//(q - 1))]),
   Parameters([Parameter(n, q^2 - 1)], [((n) * 1//(q + 1))]),
   Parameters([Parameter(n, q^2 + q + 1)], [((n) * 1//(q^2 + q + 1))]),
 ]


fingolfin:
This sounds quite plausible.

I wonder whether @frankluebeck has any thoughts on this?

frankluebeck:
I agree with your analysis that the explicitly given exceptions for class type 6 and character type 6 are not complete (information given by `PrintCharParam`/`PrintClassParam` in CHEVIE). This may be true for other class and character types, also in other tables, too. 

More precise information can be read from the corresponding character sum/ class sum functions.

In the specific case you mention you are right that the conditions `k < l < m`  (m = -k -l mod q-1) should be interpreted as "the parameters (or matrix entries) should be pairwise different", and we get any class/character in this type 6 times.

Your proposed correction would be alright if CHEVIE's (a,b) would correspond to (k,l) in the paper. But I think the chosen parameterization in this case was a = k-l and b = m-l (so a - b = k-m) such that the exceptions translate to a <> 0, b <> 0, a - b <> 0 mod q-1 (the first two are missing in the parameter description). Compare with the corresponding class/character sum function.

SoongNoonien:
> Your proposed correction would be alright if CHEVIE's (a,b) would correspond to (k,l) in the paper. But I think the chosen parameterization in this case was a = k-l and b = m-l (so a - b = k-m) such that the exceptions translate to a <> 0, b <> 0, a - b <> 0 mod q-1 (the first two are missing in the parameter description). Compare with the corresponding class/character sum function.

Ok, but now I'm slightly confused how this guarantees that a+2b is never divisible by q-1. What about, for example, q=8, b=4, a=6? In this case, a+2b=14=2*(q-1) but neither a-b=2 nor a,b are divisible by q-1. So, this would be a problem in the decomposition, or am I not getting something right?

frankluebeck:
Another small remark: In the "extended example" it should be
```
julia> T2 = set_congruence(T; remainder=2, modulus=6);
```
instead of `0 mod 2`, because the table is not valid for `q = 1 mod 3`.

SoongNoonien:
Yes, that is correct. But the congruence is only used to simplify the results further. In this case it is sufficient to just assume divisibility by two.

SoongNoonien:
I've taken a closer look at the original table and fail to see how a=k-l and b=m-l makes sense. Take a look at the value of character 4 at class 6:
Bildschirmfoto vom 2026-05-21 16-03-36

The parameter $u$ should correspond to `n` in GenericCharacterTables.jl/CHEVIE and $\epsilon$ corresponds to what GenericCharacterTables.jl outputs as `E(q - 1)`. The original value perfectly matches the value given by GenericCharacterTables.jl/CHEVIE if we assume a=k and b=l (hence m=-a-b):
```
E(q - 1)^(a*n) + E(q - 1)^(b*n) + E(q - 1)^(-a*n - b*n)
```

But if we assume a=k-l and b=m-l, then this expression becomes:
```
E(q - 1)^(n*k - n*l) + E(q - 1)^(-m*n - n*k + 2*n*l) + E(q - 1)^(m*n - n*l)
```
And if we now plug in k=1, l=2, m=q-4 and n=1, we get:
```
E(q - 1)^-9 + E(q - 1)^5 + E(q - 1)^4
```

According to the original paper this should be:
```
E(q - 1) + E(q - 1)^-3 + E(q - 1)^2
```

And for q=8 they are really distinct. The first gives $2\zeta_7^5+\zeta_7^4$, while the original paper gives $\zeta_7^4+\zeta_7^2+\zeta_7$, where $\zeta_7 = \exp(2\pi i \frac{1}{7})$.

frankluebeck:
I have also looked at the expressions for the character values (and e.g. specialized the cases for some specific q). I agree that this does not fit together. There are many possible parameterizations, but it seems indeed that here the parameters occuring in the values correspond to the parameterization you have in mind. And the class/character sum functions implement what I have said above.

My conclusion is that the current version of this generic table is not correct.

"In principle" the table is easy to compute: All characters are restrictions of characters of GL(3,q). So one has to describe class parameters in terms of those for GL3 for the classes in SL3, and one needs character parameters which yield the pairwise different restrictions to SL3. This must be implemented in the class and character sum functions. But doing this is somewhat technical and error prone.

I have a program that computes this table automatically. But in the moment I don't want to spend much time on transforming this into a publishable format.

frankluebeck:
Concerning the `set_congruence` call in the example: One should only specify congruence conditions which are refinements of the conditions of the table in the argument. Anything else is asking for trouble and will likely lead to errors.

What you are doing in the specific example probably "accidentally" returns the correct values  (if the table were correct), but doing the same with the "SL3.1" table will certainly produce errors.

In an instructional example you should definitely follow the rule to only use refining congruence conditions.

SoongNoonien:
> And the class/character sum functions implement what I have said above.

As far as I understand the class sum function for class 6, the only difference to the parameterization I have in mind is that some symmetry in the input value `tt` is used to simplify the computation. Just for reference, this is the code:
```julia
function (tt::Union{GenericCyclo,GenericCycloFrac})
  ss5 = nesum(tt, a, 0, q - 2)
  ss6 = nesum(ss5, b, 0, q - 2)
  tt1 = evaluate(tt, [b], [a])
  ss7 = nesum(tt1, a, 0, q - 2)
  tt2 = evaluate(tt1, [a], [0])
  1//6 * ss6 - 1//2 * ss7 + 1//3 * tt2
end
```

First `tt1` fixes a=b (i.e. a-b divisible by q-1), then `ss7` sums over all possible values for a. In theory we should only subtract this once, but we actually subtract it three times. This is because the other two this correspond to the other two exceptions I proposed. The exception a+2b divisible by q-1 corresponds to a=-2b. And in fact, if we replace a by -2b and the sum over all possible values of b we get exactly `ss7`. The same holds for the exception b+2a divisible by q-1. This is because replacing b by a or by -2a yields exactly the same result if `tt` is a value of any character on class 6. The value of `tt2` corresponds to the case where k=l=m, which we subtracted three times with `ss7`. But we only need to subtract this once, so `tt2` is added back twice.

SoongNoonien:
Concerning the congruences: I'm not sure I understand why this should be a problem, even in the case of "SL3.1". The only thing `set_congruence(T; remainder=0, modulus=2)` does is assuming, from now on, that q is divisible by 2 when simplifying generic cyclotomic numbers. Then the results are guaranteed to be correct for all even q. In particular, this includes all q congruent to 2 modulo 6 in the case of "SL3.n1" or q congruent to 4 modulo 6 in the case of "SL3.1".
Am I missing something here?

I agree that it is more elegant to specify q to be congruent to 2 modulo 6 in the extended example, but note that the text is based on [CHEVIE—a system for computing and processing generic character tables](https://doi.org/10.1007/BF01190329), which also just specifies q to be congruent to 0 modulo 2 in this case. See page 202.

frankluebeck:
Looking at the example
```
julia> S = generic_character_table("SL3.1")
Generic character table SL3.1
  of order q^8 - q^6 - q^5 + q^3
  restricted to q congruent to 1 modulo 3
  with 17 irreducible character types
  with 10 class types
  with parameters (a, b, m, n)
 
julia> S2 = set_congruence(S,remainder=0,modulus=2)
Generic character table SL3.1*
  of order q^8 - q^6 - q^5 + q^3
  restricted to q congruent to 4 modulo 6
  with 17 irreducible character types
  with 10 class types
  with parameters (a, b, m, n)
```
I can see now that `set_congruence` works differently in this package than in the original CHEVIE package; actually it _refines_ the conditions stored in the argument. In that case  the example is alright.

How can I access the stored congruence conditions in a generic table? It must be stored somehow since it is shown with `print`.

Also for the summation functions: How can I see their code? The class sums seem to be in `S.classsums`, but printing does not show useful information:
```
julia> print(S.classsums[5])
#19
```
And where are the character sums?
Similarly, there is `S.chars` but where is `S.classes`?

SoongNoonien:
> I can see now that `set_congruence` works differently in this package than in the original CHEVIE package; actually it refines the conditions stored in the argument. In that case the example is alright.

Yes, that is correct. But the actual simplification works exactly the same. That's why I fail to see why a result which is correct for all even q can be wrong for some q which is congruent to 2 modulo 6 (in the case of SL3.n1). The worst "problem" I can see is that a result might look more complicated as it actually is, because it hasn't been simplified as much as it could have been.

> How can I access the stored congruence conditions in a generic table?

Actually, the congruence condition is not directly an attribute of the table. It is a property of the underlying generic cyclotomic ring. It can by accessed through the `congruence` field of the ring. So, I your case `S.ring.congruence` but both are internal fields which shouldn't be necessary to access in normal use.

> Also for the summation functions: How can I see their code? The class sums seem to be in `S.classsums`, but printing does not show useful information:

<del>That is because in `S.classusms` there are only the functions in executable form and not actual code. There might be a way to access the code somehow, similar to the `less` macro for other methods. But I'm not aware of such a feature, perhaps @fingolfin knows more about this.</del> The code of all tables is stored in `data/Tables`. So, the code for "SL3.1" is in `data/Tables/A2/SL3.1.jl`. The hierarchy is identical to that in CHEVIE.

> And where are the character sums?

They are stored in the characters themselves. So, for example in `S[6].sum`.

>  Similarly, there is `S.chars` but where is `S.classes`?

Yes, internally the table is mostly just an array of characters. A class object is created on demand. No data will be copied for that, the new object just holds references to the data in the table. The shorthand syntax for this is `S[:,5]`. More details can be found in this [section](https://oscar-system.github.io/GenericCharacterTables.jl/dev/classes/) in the docs.

frankluebeck:
How can you print that class sum function for class type 6 that is shown above?
Assuming that the parameters are the same as in the table for GL3 it looks ok (because there is an S_3 symmetry in the three class parameters).

For my comment above I had looked at the character sum for character type 6 in CHEVIE:
```
> print(CharA2202Summe6);
proc(tt)
local ss5, ss6, ss7, tt1;
    ss5 := nesum(tt, nN = 1 .. q - 2);
    ss6 := nesum(ss5, mM = 1 .. q - 2);
    tt1 := eesubs(mM = nN, tt);
    ss7 := nesum(tt1, nN = 1 .. q - 2);
    linkomb(1/6, ss6, -1/6, ss7)
end proc
``` 
This uses mM and nN in the range 0..q-2 with the exceptions nN = 0, mM = 0, nN = mM (mod q-1).

Let's look at the character value on the 6th class type and character type in the GL3  case.  Say, a, b, c are the class parameters and m, n, l are the character parameters. The character value is a sum of 6 powers of E(q-1) with exponent:
```
a m + bn + cl
```
and all permutations of {a,b,c} in that expression. 
If we now restrict to SL3 and set c = -a -b, m' = m - l, n' = n - l the exponent above becomes:
``` a m' + b n'```
and similarly for the other summands. Do we get exactly the terms given in the generic table SL3.n1?

Conclusion: we were both "half right" in the previous discussion;  for the class parameters the missing exceptions are `a + 2b = 0` and `2a + b = 0` and for the character parameters the missiong exceptions are `m = 0` and `n = 0` (all mod `q-1`).

**Comment by SoongNoonien** [comment]
> How can you print that class sum function for class type 6 that is shown above?

I just copied it from [`data/Tables/A2/SL3.n1.jl`](https://github.com/oscar-system/GenericCharacterTables.jl/blob/b0e78e008fa3e3367674dfe8fd7a1041856d78af/data/Tables/A2/SL3.n1.jl#L181).

The sum function of character 6 looks the same as in CHEVIE:
```julia
function (tt::Union{GenericCyclo,GenericCycloFrac})
  ss5 = nesum(tt, n, 1, q - 2)
  ss6 = nesum(ss5, m, 1, q - 2)
  tt1 = evaluate(tt, [m], [n])
  ss7 = nesum(tt1, n, 1, q - 2)
  1//6 * ss6 - 1//6 * ss7
end
```

> Do we get exactly the terms given in the generic table SL3.n1?

Yes, we do indeed. I've not introduced new parameters `m'` and `n'` and instead reused `m` and `n` directly:
```
julia> G=generic_character_table("GL3")
Generic character table GL3
  of order q^9 - q^8 - q^7 + q^5 + q^4 - q^3
  with 8 irreducible character types
  with 8 class types
  with parameters (a, b, c, l, m, n)

julia> S=generic_character_table("SL3.n1")
Generic character table SL3.n1
  of order q^8 - q^6 - q^5 + q^3
  with 8 irreducible character types
  with 8 class types
  with parameters (a, b, m, n)

julia> G[6,6]
E(q - 1)^(a*m + b*l + c*n) + E(q - 1)^(a*n + b*l + c*m) + E(q - 1)^(a*m + b*n + c*l) + E(q - 1)^(a*n + b*m + c*l) + E(q - 1)^(a*l + b*n + c*m) + E(q - 1)^(a*l + b*m + c*n)

julia> q,(a,b,c,l,m,n)=parameters(G);

julia> evaluate(G[6,6], [c,m,n], [-a-b,m+l,n+l])
E(q - 1)^(-a*m + a*n - b*m) + E(q - 1)^(a*m + b*n) + E(q - 1)^(a*n + b*m) + E(q - 1)^(-a*m - b*m + b*n) + E(q - 1)^(-a*n + b*m - b*n) + E(q - 1)^(a*m - a*n - b*n)

julia> S[6,6]
E(q - 1)^(-a*m - b*m + b*n) + E(q - 1)^(a*m - a*n - b*n) + E(q - 1)^(-a*n + b*m - b*n) + E(q - 1)^(-a*m + a*n - b*m) + E(q - 1)^(a*m + b*n) + E(q - 1)^(a*n + b*m)

```

The different order of the terms is due to the underlying rings being slightly different and that the terms are stored in dictionaries.

> Conclusion:...

Alright, thanks for the analysis! I'll update the PR accordingly.

SoongNoonien:
> How can you print that class sum function for class type 6 that is shown above?

Ah, I'm sorry I might have misunderstood you. You mean how to print this in the julia console? One can actually use the `less` macro as usual. I wasn't aware of that before.

```
julia> g=generic_character_table("SL3.n1")
Generic character table SL3.n1
  of order q^8 - q^6 - q^5 + q^3
  with 8 irreducible character types
  with 8 class types
  with parameters (a, b, m, n)

julia> @less g.classsums[6](g[6,6])
```

This will show the code inside your pager. Plugging in `g[6,6]` is necessary because you have to run the function to use the `less` macro. That is what I forgot before.

fingolfin:
Side remark: to view the code for `S.classsums[5]` you can use `less` (or use `edit` to open it in an editor).

Note that in general, Julia functions are roughly equivalent to GAP operations: so really what one should do is to look at a specific *method*. But since there is just one here, Julia is clever enough to assume that's what you want to look at. More general is to use `methods`:
```
julia> methods(S.classsums[5])
# 1 method for anonymous function "#19":
 [1] (::GenericCharacterTables.var"##CHAR_TABLE#277".var"#19#20")(tt::Union{GenericCharacterTables.GenericCyclo, GenericCharacterTables.GenericCycloFrac})
     @ ~/Projekte/OSCAR/GenericCharacterTables.jl/data/Tables/A2/SL3.1.jl:310
```

SoongNoonien:
I'll go ahead and merge this now. If I find similar issues in other tables I'll try to make a new PR.

PR #325, opened 2026-08-06
Improve `specialize` for characters and fix `evaluate` for exceptions

SoongNoonien:
Especially with #324 it is quite helpful to have the parameters and exceptions of characters updated when specializing. This reduces the manual work necessary to go through the remaining exceptions when computing scalar products.

### DIFF
--- a/docs/src/book.md
+++ b/docs/src/book.md
@@ -98,7 +98,6 @@ With exceptions:
   2*m1 - n1 ∈ (q - 1)ℤ
   m1 - 2*n1 ∈ (q - 1)ℤ
   m1 + n1 ∈ (q - 1)ℤ
-  m1 - n1 ∈ (q - 1)ℤ
 <7, h> = 0
 With exceptions:
   n1 ∈ (q - 1)ℤ
--- a/src/GenericCharacter.jl
+++ b/src/GenericCharacter.jl
@@ -50,7 +50,7 @@ function tensor_product(char1::GenericCharacter, char2::GenericCharacter)
   param1 = shift_char_parameters(t, char1.params, 4)
   param2 = shift_char_parameters(t, char2.params, 5)
   new_char_params = Parameters(
-    vcat(param1.params, param2.params), vcat(param1.exceptions, param2.exceptions)
+    vcat(param1.params, param2.params), ParameterExceptions(vcat(param1.exceptions.exceptions, param2.exceptions.exceptions))
   )
   return GenericCharacter(
     t,
@@ -237,7 +237,7 @@ function linear_combination(coeffs::Vector{<:RingElement}, chars::Vector{<:Gener
     params[i] = shift_char_parameters(t, chars[i].params, 5 + i)
   end
   new_char_params = Parameters(
-    vcat(map(x -> x.params, params)...), vcat(map(x -> x.exceptions, params)...)
+    vcat(map(x -> x.params, params)...), ParameterExceptions(vcat(map(x -> x.exceptions.exceptions, params)...))
   )
   return GenericCharacter(
     t,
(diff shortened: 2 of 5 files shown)


fingolfin (approved):

fingolfin (commented):
Max here: I let Claude have a look at this. It found some stuff. I'll let it report what it found below. This may contain some slop. Apologies for exposing you to it. I may try to edit it myself later, or we can discuss it next week; but *if* you are to look at it, and anything is weird and confusing, just ignore it.

---

I ran this branch against master on the GL2 table to check edge cases. The refactor (`Parameters.exceptions` as `ParameterExceptions`, iteration protocol, the `book.md` fix) looks good. The new `specialize` logic and the new `evaluate` have some problems though.

**Bugs, ordered by severity**

1. **`evaluate` on exceptions now crashes when an exception becomes zero.** The new code routes through `add_exception!`, which calls `inv(exception)`. Substituting `k -> l` in `g[3]` (exception `k - l ∈ (q-1)ℤ`), or evaluating a scalar product at `k2 = k1, l2 = l1`, dies with `0 is not invertible in Fraction field ...`. Master printed `0 ∈ ℤ`. This also affects `evaluate(::GenericCycloFrac, ...)`.

2. **Always-true exceptions are silently dropped.** `add_exception!` discards any exception that evaluates to ±1. So `specialize(g[3], k, l + q - 1)` yields `l ∈ {1,…, q - 1}` with no exception, although every value of `l` is illegal. Master kept the (ugly, but honest) exception. Once point 1 is fixed the same applies to an exception `0`: it should mean "always exceptional", not "no exception". Both are pre-existing quirks of `add_exception!` from #321, but this PR is what exposes them to `evaluate`, so `evaluate` should handle constant results (zero / integer) explicitly before calling `add_exception!`.

3. **The modulus update is only correct when the coefficient divides the modulus.** `specialize(g[1], k, 2*k)` gives `k ∈ {1,…, 1//2*q - 1//2}`. The correct new modulus is `modulus / gcd(c, modulus)`, which is not a polynomial in general and depends on the congruence class of `q`. `specialize(g[1], k, q*k)` throws `Division is not exact in divexact`. Suggestion: restrict this branch to `c == ±1` (i.e. shifts like `k -> k + 1` or `k -> -n + (q-1)x`) and otherwise record a substitution or error.

4. **`specialize(g[1], k, 1)` is now a `MethodError`.** `coeff` has no method for an `Int`. Master accepted this. Either coerce `expr` into `base_ring(t.ring)` before the `coeff` call (as is already done for the substitution), or narrow the signature.

5. **A parameter can vanish while still appearing in the values.** With `k -> k^2` the coefficient of `k^1` is zero, so `k` is removed from the parameters, no substitution is recorded, and `k` still appears in the character values. With `k -> q - 1` the parameter is removed and the substitution is lost too, so the user can no longer see what was specialised. Substitutions should probably always be recorded, not only when `var` is not a parameter.

**Minor**

- The line `push!(params.params, Parameter(var, divexact(...)))` is the one Codecov flags as uncovered. The nonzero-coefficient branch has no test at all, and it is the branch with the questionable mathematics. Tests for `k -> 2k`, `k -> -k`, `k -> l + 1`, and the crash cases above would help.
- The `specialize` docstring does not mention that parameters and exceptions are now updated.
- `tensor_product` and `linear_combination` concatenate raw exception vectors, bypassing the deduplication `merge` would give. Not wrong, just inconsistent with the new `evaluate`.
- `var_shift(::Parameters, ...)` broadcasts over a `ParameterExceptions` and relies on the `collect` fallback of `Base.broadcastable`. Works because `eltype`/`length` are defined, but a comprehension would be clearer.

SoongNoonien:
These points are mostly correct but I don't think the all of the proposed fixes are what should be done here. The second bug is the most important in my opinion. In general, I think that "always true" exceptions should raise an exception (a Julia one). I these cases the computation is always wrong and it is not possible to work through the other exceptions to fix them.

I'll make an addition PR which raises an exception (a Julia one) in `add_exception!` when an "always true" exception is passed as this is a problem everywhere `add_exception!` is called, especially in `nesum`.

